PR #277, opened 2023-09-19
Cost estimation in preview with link to faucet

### LINKED ISSUE #53: Add faucet and cost preview in final confirmation step
For now, just a text showing the number of tokens you have, as informative. In a near future we'll have to properly calculate the cost of the election, and determine if the account has enough tokens.

If the account doesn't have enough tokens, add a button for the facuet (only for dev & stg environments). In the future, for prod environment, we'll define and create a specific flow, so organizations can contact us to get tokens.

elboletaire:
Note that by having the faucet externaly, we must find a way to update the balance in the scaffold side (or try to import the faucet somehow)

closes #53


nigeon:
Faucet frontend is not external anymore.
Available at /#/faucet and the main buttons/logic for interacting with it can be loaded anywhere via `<Claim />`

Because of the oAuth flow a simple "draft" feature has been implemented, letting anyone load the last process that was in creation (by using the "loadDraft" url param).

@elboletaire not sure what you think about it, but maybe we should clean the url after all those redirects.

src/components/Faucet/Claim.tsx
```diff
@@ -72,6 +72,13 @@ export const Claim = () => {
         status: 'success',
         duration: 4000,
       })
+
+      // cleanup params from url
+      const url = new URL(window.location.href)
+      window.history.replaceState({}, '', `/${url.hash}`)
+
+      // and update stored balance
+      await fetchBalance()
```
elboletaire, line 81: This isn't working as expected. I'll check tomorrow.

PR #302, opened 2023-10-04
Faucet, token cost estimation preview and other changes

### LINKED ISSUE #278: Integrate the "Faucet" into the ui-scaffold
As Vocdoni we want to have a unified flow inside the UI-scaffold to facilitate the user-flow of requesting tokens. To do so, we will move the "faucet" inside the UI-scaffold:

1 - Having a page dedicated to the faucet, in "/faucet" with the current flow
2 - Create a component to integrate a popup in the "Create Election" flow, and be able to display the faucet and request tokens in the same flow, taking into account that in that step, the organizer is has the wallet connected and only needs to do the 2nd part of the flow (sign with the social account), receive the tokens and update it in the UI.

elboletaire:
In this deploy:
- switch to questions tab when voting
- faucet (available via `/faucet`, as directly integrated in the create process form too) closes #278 
- token cost estimation preview in the process confirm/preview page
- added plausible analytics script (cookies free/privacy friendly)
- disable vote overwrite setting when anonymous is enabled, and vice-versa.


elboletaire:
something's odd with git history... I'm wondering why is it broken now that I've been one week out 🙄

PR #319, opened 2023-10-06
revamp org creation

gerouvi:
New organization creation view following the required styles.

closes # 315


src/components/Account/Create.tsx
```diff
@@ -52,40 +62,99 @@ export const AccountCreate = () => {
...
+
+        <FormControl>
+          <Textarea
+            {...register('description')}
+            placeholder={t('form.account_create.description_placeholder').toString()}
+          />
           <FormHelperText>
             <InfoOutlineIcon />
-            <Text>{t('form.account_create.title_helper')}</Text>
+            <Text> {t('form.account_create.description_helper')}</Text>
```
elboletaire, line 106: Why add an empty space before the translation? Give it a margin, padding, or add the space in the translation itself (think how should be that space if the reading was RTL).

elboletaire (changes requested):

elboletaire (approved):

PR #331, opened 2023-10-11
new process aside divided in desktop and mobile

gerouvi:
New process, separated into desktop and mobile in the process view


elboletaire (changes requested):
The `identify` button for spreadsheet strategy processes is missing in mobile devices. 

Also, this could have probably been made using CSS grids, avoiding the need to create different components for each view.

elboletaire (approved):

PR #394, opened 2023-11-09
Added FB and Ggoogle. Moving params to base64 encoded "state"

nigeon:
**ATTENTION! THIS MUST BE ACCEPTED/DEPLOYED AT THE SAME TIME AS https://github.com/vocdoni/vocfaucet/pull/10**
Google and Facebook faucet providers buttons.

### DIFF
--- a/src/components/Faucet/Claim.tsx
+++ b/src/components/Faucet/Claim.tsx
@@ -3,10 +3,14 @@ import { ConnectButton } from '@rainbow-me/rainbowkit'
 import { errorToString, useClient } from '@vocdoni/react-providers'
 import { useEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
-import { FaGithub } from 'react-icons/fa'
-import { useFaucet } from './use-faucet'
+import { FaFacebook, FaGithub, FaGoogle } from 'react-icons/fa'
+import { signinUrlParams, useFaucet } from './use-faucet'
 
-export const Claim = () => {
+type ClaimProps = {
+  signinUrlParams?: signinUrlParams[]
+}
+
+export const Claim = (props: ClaimProps) => {
   const { t } = useTranslation()
   const { connected } = useClient()
   const { loading, handleSignIn } = useClaim()
@@ -20,10 +24,32 @@ export const Claim = () => {
             w='full'
             isLoading={loading}
             colorScheme='primary'
-            onClick={() => handleSignIn('github')}
+            onClick={() => handleSignIn('github', props.signinUrlParams || [])}
           >
             <Icon mr={2} as={FaGithub} />
             {t('login.github')}
+          </Button>
+
+          <Button
+            type='submit'
+            w='full'
+            isLoading={loading}
+            colorScheme='facebook'
+            onClick={() => handleSignIn('facebook', props.signinUrlParams || [])}
+          >
+            <Icon mr={2} as={FaFacebook} />
+            {t('login.facebook')}
+          </Button>
+
+          <Button
+            type='submit'
+            w='full'
+            isLoading={loading}
+            colorScheme='red'
+            onClick={() => handleSignIn('google', props.signinUrlParams || [])}
+          >
+            <Icon mr={2} as={FaGoogle} />
+            {t('login.google')}
           </Button>
         </Flex>
       )}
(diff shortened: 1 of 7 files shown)


src/components/Faucet/Claim.tsx
```diff
@@ -1,34 +1,128 @@
...
+    try {
+      accountInfo = await client.accountService.fetchAccountInfo(address)
+    } catch (error) {}
+
+    return !!accountInfo
+  }
+
+  const auth = (provider: string) => {
+    return handleSignIn(provider, recipientAddress, props.signinUrlParams || [])
+  }
```
elboletaire, line 60: Considering `recipientAddress` is a state value, it woulud be safer to define it using `useCallback`:


```suggestion
  const auth = useCallback((provider: string) => 
    handleSignIn(provider, recipientAddress, props.signinUrlParams || [])
  , [recipientAddress])
```

src/components/Faucet/Claim.tsx
```diff
@@ -1,34 +1,128 @@
...
+
+  const isValidAddress = async (address: string) => {
+    if (!ethers.utils.isAddress(address)) return false
+
+    let accountInfo
+    try {
+      accountInfo = await client.accountService.fetchAccountInfo(address)
+    } catch (error) {}
+
+    return !!accountInfo
```
elboletaire, line 55:
```suggestion
    try {
      return !!(await client.accountService.fetchAccountInfo(address))
    } catch (error) {}
```

src/components/Faucet/Claim.tsx
```diff
@@ -1,34 +1,128 @@
...
+  } = useForm({
+    defaultValues: {
+      address: '',
+    },
+    mode: 'onChange',
+  })
+
+  useEffect(() => {
+    setRecipientAddress(account?.address as string)
+  }, [account])
```
elboletaire, line 45: Is there any reason to specify the full `account` object? Better specify `account?.address`, since passing the entire object could cause undesired effect executions, thus causing unnecessary renders.

src/components/Faucet/use-faucet.tsx
```diff
@@ -1,7 +1,18 @@
...
+  oauth?: number
+  open?: number
+  aragondao?: number
+}
+
 export const useFaucet = () => {
-  const { connected, signer, client } = useClient()
+  const { client } = useClient()
 
   // TODO: Remove this when the client is updated
```
elboletaire, line 17: Since you're changing this file quite a lot, could you please make these changes? Theoretically, the latest SDK versions should not return v2 already, so it should be only remove the "url.replace" line below (but please test it :__))

src/components/ProcessCreate/CostPreview.tsx
```diff
@@ -26,6 +26,7 @@ import { TbDatabaseExclamation } from 'react-icons/tb'
 import { useClaim } from '~components/Faucet/Claim'
 import { useProcessCreationSteps } from './Steps/use-steps'
 import imageHeader from '/assets/voc-tokens.jpg'
+import { signinUrlParams } from '~components/Faucet/use-faucet'
```
elboletaire, line 29: Should be properly sorted. Recommended to use the sort on save option of vscode:

~~~json
"editor.formatOnSave": true,
"editor.defaultFormatter": "esbenp.prettier-vscode",
"editor.codeActionsOnSave": {
  "source.organizeImports": true
}
~~~

src/components/Faucet/Claim.tsx
```diff
@@ -54,20 +148,27 @@ export const useClaim = () => {
...
 
+    // Remove the querystring from the browser current url
+    window.history.replaceState({}, '', `${window.location.pathname}${window.location.hash}`)
+
     claimTokens(provider, code, recipient)
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [pendingClaim])
 
-  const handleSignIn = async (provider: string) => {
+  const handleSignIn = async (provider: string, recipient: string, signinUrlParams: signinUrlParams[]) => {
```
elboletaire, line 168: Could you define this (and other exported methods of the hook) types and export such types? It would be a lot more useful for end developers not having to specify the entire function signature every time they use it.

elboletaire (changes requested):

src/components/Faucet/use-faucet.tsx
elboletaire: Oof, no... I'll do it. I've added this todo to the issue tracking the SDK upgrade: https://github.com/vocdoni/ui-scaffold/issues/399

elboletaire (approved):

PR #413, opened 2023-11-17
fixed issue, updated total tokens census3

gerouvi:
fixed issue when changing the chain so that the total token count above the token select updates

### DIFF
--- a/src/components/ProcessCreate/Census/Token.tsx
+++ b/src/components/ProcessCreate/Census/Token.tsx
@@ -168,15 +168,6 @@ export const CensusTokens = () => {
           options: [{ name: t('census.request_custom_token'), status: { synced: true }, type: 'request' }],
         })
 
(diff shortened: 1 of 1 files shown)


src/components/ProcessCreate/Census/Token.tsx
```diff
@@ -203,6 +194,17 @@ export const CensusTokens = () => {
...
+    const options = groupedTokens.map((token) => token.options).flat()
+    const totalOptions = options.filter((op) => {
+      if ('chainID' in op && op.chainID === ch?.chainID) {
+        return true
+      }
+      return false
+    }).length
+    setTotalTks(totalOptions)
+  }, [ch])
+
```
elboletaire, line 207: Why use an effect and not `useMemo` here? What's the reasoning?

elboletaire (changes requested):

elboletaire (approved):

PR #415, opened 2023-11-20
Anonymous flow changes

### LINKED ISSUE #355: Anonymous flows
We should adapt the flows to the new anonymous voting. 

**Non-Anonymous Voting Flows:**

Signer:  Connect -> Select Wallet -> Vote -> Confirm -> Success
Wallet:  Identify -> Fill form -> Vote -> Confirm -> Success

**New Anonymous Voting Flows:**

Signer: Connect -> Select Wallet -> Authenticate -> Vote -> Confirm -> Voting Modal -> Success
Wallet: Identify -> Fill form + password -> Vote -> Confirm -> Voting Modal -> Success

____

When we detect that the voting process is anonymous, different uses cases should be applied depending on the census type used:

### Signer Accounts (Token Voting & Wallet addresses):

With "signer accounts", the voter will encrypt the information with their private_key by signing the transaction and generating the SIK in this action, we don't need to add a "password".

**1 - Identification Phase:**
When entering for first time, all the public voting info + a "Identify" button should be visible. Clicking into the "Identify" button will show the popup for selecting the wallet connector.

**2 - Authentication Phase:** 
After connecting, a new modal appears with the "Authentication" flow:

**2.1 [Already registered ]** 
A popup will appear, with a text that explains that with anonymous voting, we need to verify you identity by signing a TX (SIK). 

SIK Authentication:
image

**2.2 [ Not Registered ]** 
A new popup text on the modal, explaining that before voting, you need to create a SIK + an account by signing 2 TX, explaining that it's free and only needs to be done once. 

SIK Register + Account Creation:
image

_[In both cases, 2.1 and 2.2, we should store the SIK in the user storage to use it while the tab is not closed.]_

3 - Show the voting page, with all the proper info on it (if he voted or no, showing the appropriate messages and buttons).

____

### Internal Wallets (Custom Data (CSV)):

With "internal wallets" we have the control over the wallet and the signatures can be done in the background without the user noticing, but the organizer can impersonate the voter as he has all the information or check what they voted. For that reason, we introduce the "password" into the flow, to add a external info that only the voter knows.

**1 - Identification Phase:**
Without "Identify", the voting info should be visible. Clicking into the "Idenfity" button will show the form popup for entering the CSV data.

**2 - Authentication Phase:**
After connecting successfully, a new button appears with a "Authenticate": 

2.1 Not registered & no password.
If the user hasn't set the password, just after identifying, we change the "form" content to this new popup with the password setting, after that, the user can vote (in the background, we also create the account):

image

2.2 Registered with password:
If the user has set previously a password, when clicking the "Authenticate" button, a new popup will appear with the password flow. Then, the user will be able to vote, re-vote or check it's vote:

image

____

### Both Flows (Signers and Wallets)

New popup while voting:
image

**TASKS**:
- [ ]  Detect if the voting is anonymous
- [ ] Detect if the voting is signer/wallet
- [ ] Add the new flow in the signer votings
     - [ ] Add the "Authenticate" button with a text explaining it 
     - [ ] Show the new signer popup
     - [ ] Pre-fetch the circuit files (workers)
     - [ ] Add the modal when voting explaining the waiting time and process 
- [ ] Add the new flow in the wallet votings
     - [ ] Add the password in the form popup
     - [ ] Pre-fetch the circuit files (workers)
     - [ ] Add the modal when voting explaining the waiting time and process
 - [ ] Make sure the voter can only view the vote after authenticate in both cases 
 - [ ] Test

### LINKED ISSUE #414: Switching accounts does not properly work when switching from a non existing one
If you vote on an election (anonymous at least), and after receiving the voteid you switch accounts, such vote id remains in there.

This issue has appeared before and was ui-component's fault, but this time is probably related to the changes in #397, when wagmi/viem replaced ethers.

### LINKED ISSUE #428: Census gets created before creating an election
During the preview cost calculation, the census is created when it should not be created until the creation of the process.

elboletaire:
closes #355 
closes #428
closes #414

Note this pull request cannot be merged without publishing a new version of ui-components first (branch `f/anonymous-flow` in there).

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "@emotion/styled": "^11.11.0",
     "@ethersproject/abstract-signer": "^5.7.0",
     "@rainbow-me/rainbowkit": "^1.1.3",
-    "@vocdoni/chakra-components": "~0.6.0",
+    "@vocdoni/chakra-components": "^0.7.0",
     "@vocdoni/rainbowkit-wallets": "^0.2.2",
     "@vocdoni/sdk": "~0.6.0",
     "chakra-react-select": "^4.6.0",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "@rainbow-me/rainbowkit": "^1.1.3",
     "@vocdoni/chakra-components": "~0.6.0",
     "@vocdoni/rainbowkit-wallets": "^0.2.2",
-    "@vocdoni/sdk": "~0.5.1",
+    "@vocdoni/sdk": "~0.6.0",
     "chakra-react-select": "^4.6.0",
     "date-fns": "^2.30.0",
     "ethers": "^5.7.2",
@@ -81,7 +81,6 @@
   },
   "packageManager": "yarn@1.22.19",
   "resolutions": {
-    "i18next": "^23.2.11",
-    "@vocdoni/sdk": "~0.5.1"
+    "i18next": "^23.2.11"
   }
 }
--- a/src/components/Account/Create.tsx
+++ b/src/components/Account/Create.tsx
@@ -3,6 +3,7 @@ import {
   Alert,
   AlertIcon,
   Box,
+  Button,
   Flex,
   FormControl,
   FormErrorMessage,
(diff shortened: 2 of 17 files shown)


elboletaire:
ooops... I obviously did not create this from the proper branch. I'll fix it tomorrow.

PR #418, opened 2023-11-21
Upgraded rainbowkit-wallets

nigeon:
Upgraded rainbowkit-wallets
Moved FAUCET_AMOUNT to dynamic values coming from the faucet.


elboletaire (approved):
Since we're using the `getAuthTypes` effect in multiple places, I was considering adding it directly to the `useFaucet` hook, but because it's not directly used in some places, it would make unnecessary calls...

We should consider creating a context for the faucet and move some logic there (like these `useEffect`), so it can be requested once, and used everywhere without repeating code nor calls.

PR #420, opened 2023-11-21
prevent bug census 3 select request token option

gerouvi:
Double-check that if the select token option is set to 'request,' do not render the token preview or value in the select

### DIFF
--- a/src/components/ProcessCreate/Census/Token.tsx
+++ b/src/components/ProcessCreate/Census/Token.tsx
@@ -300,7 +300,9 @@ export const CensusTokens = () => {
               else return `${props.name}`
             }}
             onChange={(token) => {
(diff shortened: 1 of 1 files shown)


src/components/ProcessCreate/Census/Token.tsx
```diff
@@ -300,7 +300,9 @@ export const CensusTokens = () => {
               else return `${props.name}`
             }}
             onChange={(token) => {
-              setValue('censusToken', token || undefined)
+              console.log(token.type === 'request')
```
elboletaire, line 303: duude 😅

elboletaire (changes requested):

elboletaire (approved):

PR #429, opened 2023-11-27
Implements Process authentication through a URL that contains the privKey

emmdim:
Currently it is not restricted to csv censuses.
Not sure if it should be though....
WDYT @jpaulet @elboletaire ?


elboletaire:
I'd say defining a route is a bad idea, since we're using the browser router (and with this implementation, all private keys would be logged to any backend serving the app). 

I'd say to use the hashed part of the url, no need for any key btw, just have the entire private key after the `#` in the url.

src/components/Process/Aside.tsx
```diff
@@ -2,24 +2,45 @@ import { Box, Button, Flex, Link, Spinner, Text, useMediaQuery } from '@chakra-u
...
+    async function checkPkey() {
+      if (pkey) {
+        let client = new VocdoniSDKClient({
+          env,
+          wallet: new Wallet(pkey),
+          electionId: election?.id,
+        })
+        let pkeyIsInCensus = await client.isInCensus()
+        setClient(client)
+        setPkeyAccess(pkeyIsInCensus)
```
elboletaire, line 38: Why store this value in state? Actually... there's no need to check if it's in census, the react-providers hook should do their magic and check it once the client has been set.

src/router/Router.tsx
```diff
@@ -52,6 +52,16 @@ export const RoutesProvider = () => {
...
+          <Route
+            path='processes/:id/:pkey'
+            element={
+              <SuspenseLoader>
+                <Process />
+              </SuspenseLoader>
+            }
+            loader={async ({ params }) => client.fetchElection(params.id)}
+            errorElement={<Error />}
+          />
```
elboletaire, line 64: As I said yesterday, I don't think we should be using the router to define a specific route, since doing it this way, all private keys will be logged in the backend servers where this app is deployed. I'd say to use the hash part of the route.

elboletaire (changes requested):

src/components/Process/Aside.tsx
elboletaire: BTW why in Process/Aside.tsx? Doesn't look like the easiest way to find this logic. I'd say avoid having this kind of logic here tbh.. it would be better in Process/View.tsx IMO

PR #451, opened 2023-12-04
minor fixups to english copy

imw:


elboletaire (approved):
Thanks for the changes! As non-native English speakers, we truly appreciate these changes 😃

PR #454, opened 2023-12-11
fixed issues aside buttons

gerouvi:
Fixed issues in the aside buttons, aside footer mobile, and logout


elboletaire (changes requested):
Since apparently you cannot style these buttons in any other way, please create a new `VoteButton` component and place all the styles in there, rather than having them copy-pasted twice.

elboletaire (approved):

Issue #450, opened 2023-12-04
Improve public voting page 


jpaulet:
As Vocdoni, we want to make easy for the voters to vote and understand easily the UI/layout. After some feedback, we want to improve the public voting page, by:

- [x] Changing "ongoing" text to "active"
- [x] Adding the "State" text next to the "state" label
- [x] Increase the "State" label padding and size. 
- [x] Change the "active" label to dark green + white text.
- [x] Read more background
- [x] Add the "Schedule" text next to the start/end date. 
- [x] Make the start_end date font black.
- [x] Reduce the right-side info [ends,privacy,census...)  opacity to 0.85 (on-hover, back to 1). 
- [x] Align horizontally the "< Back to org" label with the header+title. Make the "<" image 5px width.

Split the layout into 2 columns, then, adapt it to the screenshot:
- [x] Remove the description height.
- [x]  Add the questions/results component just after the description (with a space between the 2 elements)
- [x] Improve the questions buttons:
       - [ ] Add a circle check in the left side.
       - [ ] Add a "(select one)" text in the right side (very subtle)
       - [ ] Move the CTA button from the aside to be place under the questions (Connect / Sign-in first / Vote), with the description text just below it. This button, should behave like the mobile one, be always visible in the screen. 
- [x] #503

Reference screenshot:
![Screenshot 2023-12-04 at 08-37-28 The voice of digital voting](https://github.com/vocdoni/ui-scaffold/assets/6894329/9c6c05b1-7187-47b2-a825-a0640ddf297b)

PR #465, opened 2023-12-20
public voting

### LINKED ISSUE #456: Enhance Visibility of URLs in the Description fields
Problem:
Currently, clickable links appear like regular text, with the same format as the rest of the text.

Solution:
Improve the visibility of clickable links by adding a different color or a underline to the text.

gerouvi:
- added STATE and SCHEUDLE labels
- Reduced header right side
- Reduced election description line height
- closes #456

refs #450

### DIFF
--- a/src/components/Organization/View.tsx
+++ b/src/components/Organization/View.tsx
@@ -134,7 +134,7 @@ const OrganizationView = () => {
                       <Trans
                         i18nKey='organization.elections_list_empty.footer'
                         components={{
-                          customLink: <Link variant='primary' href='#' target='_blank' />,
+                          customLink: <Link href='#' target='_blank' />,
                         }}
                       />
                     </Text>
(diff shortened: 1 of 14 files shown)


src/components/Organization/View.tsx
```diff
@@ -134,7 +134,7 @@ const OrganizationView = () => {
                       <Trans
                         i18nKey='organization.elections_list_empty.footer'
                         components={{
-                          customLink: <Link variant='primary' href='#' target='_blank' />,
+                          customLink: <Link href='#' target='_blank' />,
```
elboletaire, line 137: We should properly set this link here @jpaulet

elboletaire (changes requested):
Overall looks good, but we have a link pointing nowhere that we should update first. Also, please review the Aside component links, since there they have a contrast problem

imatge

jpaulet (changes requested):
The `No description provided` is shown when there's description.

elboletaire (approved):

PR #481, opened 2024-01-04
Feature toggle

### LINKED ISSUE #426: Feature toggle
We want a feature toggle system, so deploying variations of the ui-scaffold could be defined by just an env var.

The easiest way to do this is by just not rendering components but, ideally, the feature toggle should be done at build level, so the resulting application does not contain any of the non-desired features' contents.

elboletaire:
closes #426


jpaulet:
Missing the `justify-content: start` in the "Census" tab. Other than that, everything works smoothly

PR #493, opened 2024-01-22
questions process radio and checkbox styles

gerouvi:
In the questions process, radio button styles for single choice, and checkbox styles for multiple choices


elboletaire (changes requested):
Looks good on the default theme. But you forgot checking it with onvote theme, and I don't see a `check-icon.png` file in that theme, so this would fail.

elboletaire (approved):
My fault, I wasn't aware you were using different icons and it was already defined for onvote (so you were reusing code there).

PR #526, opened 2024-01-30
F/census csp google

nigeon:
It's possible to create a census using the csp with google accounts.
This feature is dependant on https://github.com/vocdoni/ui-components/pull/138


src/components/ProcessCreate/StepForm/Censuses.tsx
```diff
@@ -44,15 +44,13 @@ export const TokenCensus = () => (
...
-export const CspCensus = () => (
-  <SuspenseLoader>
-    <Csp />
-  </SuspenseLoader>
+export const CspCensusGithub = () => (
+  <SuspenseLoader>{import.meta.env.features._census.csp_github && <Csp provider='github' />}</SuspenseLoader>
+)
+
+export const CspCensusGoogle = () => (
+  <SuspenseLoader>{import.meta.env.features._census.csp_google && <Csp provider='google' />}</SuspenseLoader>
```
elboletaire, line 55: This does not work for the treeshake... Sadly it seems the conditionals can only be defined as in the components above (in a clear `if` statement condition). Ternary statements, at least inside a react component view, do not seem to be working.

src/components/ProcessCreate/Confirm/CensusCspGithub.tsx
```diff
@@ -0,0 +1,54 @@
...
+          children={t('form.process_create.confirm.census_preview_is_shortened', {
+            limit: CensusPreviewRowsLimit,
+          })}
+        />
+      )}
+    </Flex>
+  )
+}
+
+export default PreviewCensusCspGithub
```
elboletaire, line 54: Considering how equal these two components are (and I bet they're quite close to some other component, like the web3 addresses one), it would be better to create a generic component, and define containers for each implementation.

elboletaire (changes requested):

src/components/ProcessCreate/Confirm/CensusCspGithub.tsx
nigeon: There's 2 components that are really similar (GoogleUsers and Web3) will check if I can refactor to just use 1 component

elboletaire: What about the web3 addresses one, wasn't quite similar too?

nigeon: that's basically what I've said in my comment

elboletaire: missread, my fault 🙃

elboletaire:
I'm closing this for obvious reasons. Probably shall be used for reference if we end up adding this.

PR #527, opened 2024-01-31
rounded census

gerouvi:
- when advanced settings anonymous is true show disclaimer text
- show anonymous census blance accuracy in preview

### DIFF
--- a/src/i18n/locales/ca.json
+++ b/src/i18n/locales/ca.json
@@ -500,7 +500,7 @@
   },
   "process_create": {
     "anonymous": {
-      "legal_disclaimer": "Aquest sistema de votació utilitza el poder de votació basat en tokens (públic a la cadena de blocs). Com heu seleccionat l'opció de 'Votació anònima', els pesos es gruparan en saldos similars i es redoniran cap avall. Això sacrifica una mica de precisió, però preserva la identitat anònima desvinculant els saldos de les carteres.",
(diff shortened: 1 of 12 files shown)


elboletaire (changes requested):
Found some typos, but also... I'm not convinced in the way this is being displayed @jpaulet.

I would not add the disclaimer here:
imatge

But here makes a lot of sense, and it isn't added there:
imatge

jpaulet:
> Found some typos, but also... I'm not convinced in the way this is being displayed @jpaulet.
> 
> I would not add the disclaimer here: imatge
> 
> But here makes a lot of sense, and it isn't added there: imatge

You are right, @gerouvi can you move this text to the "census" step. Use the default Chakra component for "Info Alert", place it here: 
image

With the text: `This voting system utilizes token-based voting power (public on-chain). As you selected the "Anonymous voting" option, weights will be grouped into similar balances and rounded down. This sacrifices some accuracy but preserves anonymous identity by unlinking balances from wallets.`

It will only appear if "anonymous voting" option was selected in the first step.

src/i18n/locales/ca.json
```diff
@@ -500,7 +500,7 @@
   },
   "process_create": {
     "anonymous": {
-      "legal_disclaimer": "Si els votants utilitzen el poder de vot del token (públic a la cadena de blocs), per millorar l'anonimat, els pesos s'agrupen en saldos similars, arrodonits cap avall. Això sacrifica una certa precisió però preserva la identitat desvinculant els saldos de les carteres.",
+      "legal_disclaimer": "Aquest sistema de votació utilitza el poder de votació basat en tokens (públic a la cadena de blocs). Com heu seleccionat l'opció de 'Votació anònima', els pesos es gruparan en saldos similars i es redoniran cap avall. Això sacrifica una mica de precisió, però preserva la identitat anònima desvinculant els saldos de les carteres.",
```
elboletaire, line 503: Should be "s'agruparan" and "i s'arrodoniran"

elboletaire (changes requested):

elboletaire (approved):

PR #534, opened 2024-02-01
md headings and process sizes

gerouvi:
Created 'md-sizes' class to modify the size of h1, h2, h2, and p in the Markdown renderings on the process view description and question. Adjusted process title size to align with Markdown sizes


elboletaire (changes requested):
Looks good, but any `:first-of-type` children in this new defined class should not have a top margin, otherwise it appears weird both in the preview and in the final result:

imatge

elboletaire (approved):

PR #557, opened 2024-02-22
header image and video

gerouvi:
If header is defined in the metadata, display the image, and if streamUri is defined, display the video/stream. When scrolling, position it in the top right


jpaulet:
We should add more vertical spacing between the streaming with the other elements, add a margin-top (40px) & margin-bottom (50px).

Also, when is a float element, it get's overlapped with the header buttons. Try to solve it by moving the element down.

jpaulet (approved):

elboletaire (approved):

PR #558, opened 2024-02-22
pxll theme

gerouvi:
pxll theme with brand colors


src/components/ProcessCreate/Confirm/Preview.tsx
```diff
@@ -22,6 +22,8 @@ const Preview = () => {
     endDate: end,
   } = form
 
+  console.log(description)
```
elboletaire, line 25: 👀

elboletaire (changes requested):

elboletaire:
Converted to draft since this PR should not be merged finally. Instead, we'll rename the branch to `d/pxll` and we'll deploy that directly.

PR #594, opened 2024-03-14
Fix gitcoin census creation time

selankon:
Use the newly created strategy instead of the GPS strategy

### DIFF
--- a/src/components/ProcessCreate/Steps/Confirm.tsx
+++ b/src/components/ProcessCreate/Steps/Confirm.tsx
@@ -343,14 +343,18 @@ const getCensus = async (env: EnvOptions, form: StepsFormValues, salt: string) =
       })
 
       const retryTime = 5000
-      const attempts = form.timeToCreateCensus / retryTime
-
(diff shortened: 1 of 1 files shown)


src/components/ProcessCreate/Steps/Confirm.tsx
```diff
@@ -343,14 +343,18 @@ const getCensus = async (env: EnvOptions, form: StepsFormValues, salt: string) =
       })
 
       const retryTime = 5000
-      const attempts = form.timeToCreateCensus / retryTime
-
       c3client.queueWait.retryTime = retryTime
-      // clamp attempts between 20 and 100
-      c3client.queueWait.attempts = Math.min(Math.max(Math.ceil(attempts), 20), 100)
```
emmdim, line 350: @selankon  I would not remove the above lines and I would just update the corresponding values in the case of gitcoin passport

selankon: 67825c3

emmdim (approved):
LGTM

Issue #588, opened 2024-03-13
Can't read vote text button after hovering over it


elboletaire:
![imatge](https://github.com/vocdoni/ui-scaffold/assets/153305/8b372e3b-66ee-4ca1-acee-8b6a6afc3256)

selankon:
lol this is not happening to me...

Where is this button? 

![image](https://github.com/vocdoni/ui-scaffold/assets/16777076/a386f62e-ce02-4c15-903f-759ea8128d67)

selankon:
Ok I saw that this styles are applied at the moment to vote, but it don't show the button loader (not the vote text..). Have to be fixed there at least

selankon:
This is defined here

https://github.com/vocdoni/ui-scaffold/blob/1da1fc205d21ccd45a285cf7bd364cc7ed3118d6/src/components/Process/VoteButton.tsx#L3-L10

And here:

https://github.com/vocdoni/ui-scaffold/blob/1da1fc205d21ccd45a285cf7bd364cc7ed3118d6/src/themes/onvote/components/Button.ts#L136-L142

We should define the color for the other states when disabled (`_before`, `_after` etc...).

selankon:
BTW the button commented on the issue description may be related to https://github.com/vocdoni/ui-scaffold/issues/587 and may be different at the one shown when voting

selankon:
Found the problem:

- At first I thought that prebolem was here https://github.com/vocdoni/ui-scaffold/blob/1da1fc205d21ccd45a285cf7bd364cc7ed3118d6/src/components/Process/VoteButton.tsx but in reality this component is not used indeed and there was a error on my IDE that points there. So I suggest to delete this file. 
- Then I found that the real vote button is in reality this one: https://github.com/vocdoni/ui-scaffold/blob/1da1fc205d21ccd45a285cf7bd364cc7ed3118d6/src/components/Process/Aside.tsx#L198 and is just needed to fix the variant

selankon:
Finally the problem was this: 

https://github.com-selankon/vocdoni/ui-scaffold/blob/1da1fc205d21ccd45a285cf7bd364cc7ed3118d6/src/components/Process/Aside.tsx#L203-L205

It seems that this attribute overrides the button disabled props defined on the theme. Changing by this fixes the problem

```ts
        sx={{
          '&::disabled': {
            opacity: '0.8',
          },
        }}
``` 

[...]

PR #600, opened 2024-03-15
Fix vote button

### LINKED ISSUE #586: Verify vote on explorer is not visible
imatge

selankon:
Fixes https://github.com/vocdoni/ui-scaffold/issues/588
Fixes #586

I putted also the fix for #586 because is little and related to styles


src/components/Process/VoteButton.tsx
elboletaire, line 1: Nice catch. I wasn't aware we stopped using this component tbh.

elboletaire (approved):

Issue #562, opened 2024-02-26
Improve Onvote census info


jpaulet:
In the voting page, we need to show the census information accordingly, if the census size and the max votes allowed are different, display them in separate rows under the "Census" information area.

![image](https://github.com/vocdoni/ui-scaffold/assets/6894329/3b3a0041-96b6-4313-ba0b-c47a28b409da)

PR #612, opened 2024-03-25
Improve census info

selankon:
Fixes https://github.com/vocdoni/ui-scaffold/issues/562

Add total census size information if the census size differs from maxCensusSize. 

image

image

It has a problem when runing `yarn translations` it add a `total_census_size_tooltip_one` key, which is not needed and nor used (this case should not going to happen). @elboletaire should we take attention to it or can we ignore it?

### DIFF
--- a/src/components/Process/Header.tsx
+++ b/src/components/Process/Header.tsx
@@ -149,8 +149,8 @@ const ProcessHeader = () => {
                 color='white'
                 placement='top'
                 label={t('process.total_census_size_tooltip', {
-                  max: censusInfo?.size,
-                  count: election?.maxCensusSize,
+                  censusSize: censusInfo?.size,
+                  maxCensusSize: election?.maxCensusSize,
                   percent:
                     censusInfo?.size && election?.maxCensusSize
                       ? Math.round((election?.maxCensusSize / censusInfo?.size) * 100)
@@ -159,8 +159,8 @@ const ProcessHeader = () => {
               >
                 <Text>
                   {t('process.total_census_size', {
-                    max: censusInfo?.size,
-                    count: election?.maxCensusSize,
+                    censusSize: censusInfo?.size,
+                    maxCensusSize: election?.maxCensusSize,
                   })}
                 </Text>
               </Tooltip>
--- a/src/components/Process/Header.tsx
+++ b/src/components/Process/Header.tsx
@@ -1,5 +1,5 @@
-import { WarningIcon } from '@chakra-ui/icons'
-import { Box, Button, Flex, Icon, Image, Text } from '@chakra-ui/react'
+import { InfoIcon, WarningIcon } from '@chakra-ui/icons'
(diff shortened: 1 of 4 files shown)


selankon:
> Right now, in some cases, the string appears incomplete, without the "total". Also, we agreed that if the `max allowed voters` >= `census size`, we won't show the message, only show: "XXXX voters".

Ok this commit should be enough 2034606

However, i found that the census info is undefined. It seems that the census3 is returning 404 with this census. 

Anyway, now if the census3 has an error it won't show the wrong string

src/components/Process/Header.tsx
```diff
@@ -30,7 +34,16 @@ const ProcessHeader = () => {
   )
   const strategy = useStrategy()
 
+  useEffect(() => {
```
elboletaire, line 37: Please try to add comments to every `useEffect`

selankon:
> we agreed that if the `max allowed voters` >= `census size`

Just to clarify... It should be `max allowed voters` < `census size`, only show the allowed voters strings if the election maxCensusSize is little than the census size. Isn't it?

src/components/Process/Header.tsx
```diff
@@ -30,7 +34,16 @@ const ProcessHeader = () => {
...
+    ;(async () => {
+      if (!election?.census?.censusId || !client) return
+      const censusInfo: CensusInfo = await client.fetchCensusInfo(election.census.censusId)
```
elboletaire, line 40: What happens if the call fails? Shouldn't this maybe be in a try/catch and handle the error somehow?

src/i18n/locales/en.json
```diff
@@ -493,6 +493,10 @@
       "text": "<p>We've securely processed and stored your vote on our Blockchain. Click <verify>here</verify> to verify your vote.</p>",
       "title": "Vote submitted!"
     },
+    "total_census_size_one": "",
+    "total_census_size_other": "{{count}} allowed voters of {{max}} total in census",
+    "total_census_size_tooltip_one": "",
+    "total_census_size_tooltip_other": "The maximum number of voters allowed is limited to {{count}} from a census of {{max}} ({{percent}}% of the total). Only the first {{count}} voters can vote.",
```
elboletaire, line 499: These translations are being generated because you're using the special word `count`. If you're not planning to use multiple translations for singular/plural, you should avoid using this, since it will fail when `count == 1`, because there's no translation defined for it.

elboletaire (changes requested):

elboletaire:
Also, try to update the branch via rebase or merge.

src/components/Process/Header.tsx
selankon: If the call fails the census info it remains just undefined so doesn't affect to the rest of the code. Anyway, i'll set this inside a try catch to avoid console errors

selankon: d4978e9

src/i18n/locales/en.json
selankon: dd9e9ce

elboletaire (approved):

PR #615, opened 2024-03-26
Implement share modal

selankon:
Fixes https://github.com/vocdoni/interoperability/issues/148


jpaulet:
1 - When I hover over any icon, it displays: "Share on [object Object]."
2 - Make the font size of the input smaller (like 12px) so that the URL looks clearer. Also, place the "Copy" button below to avoid cutting the input off.
3 - Make the share icon more visible (as the design provided)

image
image

jpaulet (approved):

selankon:
> 1 - When I hover over any icon, it displays: "Share on [object Object]."
> 2 - Make the font size of the input smaller (like 12px) so that the URL looks clearer. Also, place the "Copy" button below to avoid cutting the input off.
> 3 - Make the share icon more visible (as the design provided)

This changes are already fixed and the pr is approved by Aulet

elboletaire (changes requested):
This is what I see on the deployed instances:
imatge
imatge

selankon:
> This is what I see on the deployed instances:

Well this is so strange because is not happening to me on my FF....  Maybe is the FF cache related? Which is the problem, the icons doesn't load or have a bad style?

image

elboletaire:
There's a style hiding them:

imatge

elboletaire:
It was adblocker doing its magic, I love it
imatge

elboletaire (approved):

PR #617, opened 2024-03-26
Fixes census3 census creation attemts to 100

emmdim:


src/components/ProcessCreate/Steps/Confirm.tsx
```diff
@@ -356,7 +356,7 @@ const getCensus = async (env: EnvOptions, form: StepsFormValues, salt: string) =
         // Once strategy is created, we got the stimation again to update the awaiting time
         const { timeToCreateCensus } = await c3client.getStrategyEstimation(strategyID, form.electionType.anonymous)
         attempts = timeToCreateCensus / retryTime
```
elboletaire, line 358: If we're gonna hardcode this, remove the unused `attempts` definition (here and some lines above).

elboletaire (changes requested):

PR #582, opened 2024-03-12
2024-04 1st sprint staging upgrade

elboletaire:
Most significant changes in this staging upgrade:
- #555
- #570
- #572 
- #584 
- #596 
- #607
- #615 
- #617

PR #618, opened 2024-03-27
Fix gitcoin flow

selankon:
It fixes this comments https://github.com/vocdoni/ui-scaffold/pull/582#issuecomment-2021054234

It not implements the stamps visualization on the election view because it will be implemented on https://github.com/vocdoni/ui-scaffold/pull/626


selankon:
We could merge this and then fix this on another PR? https://github.com/vocdoni/interoperability/issues/174

elboletaire (approved):

PR #626, opened 2024-04-02
Implement show stamps

selankon:
Show stamps on the election page as suggested by @jpaulet on this comment https://github.com/vocdoni/ui-scaffold/pull/582#issuecomment-2021096267

image

It depends on https://github.com/vocdoni/interoperability/issues/174

### DIFF
--- a/src/components/Process/Header.tsx
+++ b/src/components/Process/Header.tsx
@@ -1,5 +1,5 @@
 import { InfoIcon, WarningIcon } from '@chakra-ui/icons'
-import { Box, Button, Flex, Icon, Image, SliderThumb, Text, Tooltip } from '@chakra-ui/react'
+import { Box, Button, Flex, Icon, Image, Text, Tooltip } from '@chakra-ui/react'
 import {
   ElectionDescription,
   ElectionSchedule,
@@ -8,7 +8,7 @@ import {
   OrganizationName,
 } from '@vocdoni/chakra-components'
 import { useClient, useElection, useOrganization } from '@vocdoni/react-providers'
-import { CensusType, ElectionStatus, Strategy } from '@vocdoni/sdk'
+import { Census3StrategyToken, CensusType, ElectionStatus, Strategy } from '@vocdoni/sdk'
 import { useTranslation } from 'react-i18next'
 import { Link } from 'react-router-dom'
 import { useReadMoreMarkdown } from '~components/Layout/use-read-more'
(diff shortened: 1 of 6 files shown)


src/components/Process/Header.tsx
```diff
@@ -229,6 +233,60 @@ const ProcessHeader = () => {
   )
 }
 
+// todo(kon): disable this cast when ready on the sdk
+type Census3StrategyTokenIcon = Census3StrategyToken & { iconURI: string }
```
elboletaire, line 237: Hhmm... what cast?

src/components/Process/Header.tsx
```diff
@@ -18,6 +18,7 @@ import { CreatedBy } from './CreatedBy'
 import { ProcessDate } from './Date'
 import { ShareModalButton } from '~components/Share'
 import { useEffect, useState } from 'react'
+import { StampIcon } from '~components/ProcessCreate/Census/Gitcoin/StampIcon'
```
elboletaire, line 21: Wrong import order. Still without the IDE feature to organize imports enabled? 👀

src/components/Process/Header.tsx
```diff
@@ -229,6 +233,60 @@ const ProcessHeader = () => {
...
 }
 
+// todo(kon): disable this cast when ready on the sdk
+type Census3StrategyTokenIcon = Census3StrategyToken & { iconURI: string }
+
+const GitcoinStrategyInfo = () => {
+  const { t } = useTranslation()
+  const { election } = useElection()
+
+  if (!election || (election && !election?.meta?.strategy)) return ''
```
elboletaire, line 243: Do not return empty string, simply `return` if you don't want the component to be rendered.

```suggestion
  if (!election || (election && !election?.meta?.strategy)) return
```

src/components/Process/Header.tsx
```diff
@@ -229,6 +233,60 @@ const ProcessHeader = () => {
...
+  const strategy: Strategy = election.get('strategy')
+
+  const score = strategy.tokens['GPS'].minBalance
+  const firstParenthesesMatch = strategy.predicate.match(/\(([^)]+)\)/)
+  let unionTypeString: string | null = null
+  if (firstParenthesesMatch) {
+    const unionType = firstParenthesesMatch[1].split(' ')[1] // split by space and get the second element which should be the union type
```
elboletaire, line 250:
```suggestion
    const [,unionType] = firstParenthesesMatch[1].split(' ') // split by space and get the second element which should be the union type
```

elboletaire (changes requested):
Looks good. Some minor notes though (see review).

src/components/Process/Header.tsx
elboletaire: BTW... the comment saying "should be the union type" makes me thing this code block could fail, so it should be inside a try/catch.

elboletaire (approved):

PR #634, opened 2024-04-03
Questions focus when unanswered question

gerouvi:
Place focus on the first unanswered question when voting

### DIFF
--- a/src/components/Process/View.tsx
+++ b/src/components/Process/View.tsx
@@ -44,8 +44,8 @@ export const ProcessView = () => {
   const electionRef = useRef<HTMLDivElement>(null)
   const { isConnected } = useAccount()
   const [tabIndex, setTabIndex] = useState(0)
-  const [err, setErr] = useState(0)
-  const [rerenderQs, setRerenderQs] = useState(false)
+  const [errPos, setErrPos] = useState(0)
+  const [rerender, setRerender] = useState(false)
 
   const handleTabsChange = (index: number) => {
     setTabIndex(index)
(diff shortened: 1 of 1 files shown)


src/components/Process/View.tsx
```diff
@@ -71,6 +75,39 @@ export const ProcessView = () => {
     if (election?.status === ElectionStatus.RESULTS) setTabIndex(1)
   }, [election])
 
+  useEffect(() => {
```
elboletaire, line 78: Please add comments to all `useEffects` describing its usage.

src/components/Process/View.tsx
```diff
@@ -71,6 +75,39 @@ export const ProcessView = () => {
...
+    if (!err) return
+
+    if (err === 1) return setErr((prev) => prev + 1)
```
elboletaire, line 81: Single line returns are ok for the first case, but the second one would be better in brackets

```suggestion
    if (!err) return

    if (err === 1) {
      return setErr((prev) => prev + 1)
    }
```

src/components/Process/View.tsx
```diff
@@ -71,6 +75,39 @@ export const ProcessView = () => {
...
+      })
+
+      if (elementsWithInvalidData.length) {
+        const firstElementWithInvalidData = elementsWithInvalidData[0].children[0].children[1]
+          .children[0] as HTMLElement
+        firstElementWithInvalidData.setAttribute('tabIndex', '0')
+        firstElementWithInvalidData.focus()
+      }
+    }
+  }, [err])
```
elboletaire, line 101: Considering the amount of accesses to arbitrary array index positions, all this `useEffect` block should be inside a try/catch just in case.

Edit: We gotta talk about how you solved this because I have my doubts about all this DOM traversing.

src/components/Process/View.tsx
```diff
@@ -71,6 +75,39 @@ export const ProcessView = () => {
...
+
+  useEffect(() => {
+    setRerenderQs(false)
+  }, [isConnected])
+
+  useEffect(() => {
+    setRerenderQs(true)
+  }, [rerenderQs])
```
elboletaire, line 109: As said above, always try to add comments to your useEffects describing what are they doing.

elboletaire (changes requested):
I don't think there should be that amount of DOM traversing in the code tbh. We gotta talk, so you can properly explain me why you did it this way, and see if we can improve it. All other comments are still valid though.

src/components/Process/View.tsx
```diff
@@ -71,6 +75,39 @@ export const ProcessView = () => {
     if (election?.status === ElectionStatus.RESULTS) setTabIndex(1)
```
elboletaire, line 75: Hey this condition also inside brackets please. Avoid single line conditions when doing logic.

elboletaire (approved):

Issue #641, opened 2024-04-05
The number selector in MaxCensusSizeSelector does not allow writing more than a single digit


elboletaire:
You can actually add more than one digits... but not at a time... each time you add a number you have to wait for the request to finish, because it changes `maxCensuSize`, which triggers the update (I think)

![imatge](https://github.com/vocdoni/ui-scaffold/assets/153305/c792bb53-2221-4249-9b5e-b2f18e4680c8)

PR #642, opened 2024-04-08
Fix maxCensusSize input number

selankon:
Fix https://github.com/vocdoni/ui-scaffold/issues/641


elboletaire (approved):
Works and code is ok, but it leaves me with a thorn in my side, since the gitcoin census also has an input number with a slider, and the behavior there is different. But since we're all a bit out of time, let's accept this anyway.

PR #657, opened 2024-04-18
New landing revamp

jpaulet:
Improved landing page by:

- Added new images
- Improved texts
- Increased and unified font-size
- Added "basier" font and font-family
- Added new layout in Benefits, Features, Clients and Contact-us
- Added background image
- Modified clients icons and removed "Pirates"
- Added new styles and some background textures & figures
- Unified margins and paddings between sections
- Other minor improvements and unifications


public/default/assets/decidim.png
elboletaire, line 1: Even though the background is transparent, it may have a thin layer of gray color, because it shows gray in the final result:

imatge

elboletaire (changes requested):
Other than the issue with the logo background, I'm not sure I like the buttons having an underline text, which is only seen here in the landing.

imatge

public/default/assets/decidim.png
jpaulet: It's ok. I will upload a new one later if necessary.

jpaulet: Also, the button with underline is to give more emphasis to the CTA. Go ahead. 

I will review it in the next iterations but this shouldn't block the merge.

elboletaire (approved):

PR #665, opened 2024-04-25
Refactor readme using template

NateWilliams2:


elboletaire (changes requested):
LGTM. Although I realized we should also add links to each onvote deployment in the `Branching and deploys` section. Could you do it in this PR?

https://dev.onvote.app
https://stg.onvote.app
https://onvote.app

elboletaire (approved):

Issue #661, opened 2024-04-24
Visual Improvements 


jpaulet:
We should address the following visual improvements:

- [x] Improve OR/AND selector in the "Gitcoin Passport" census type. Change the current switch to the "custom radio button" Chakra component, while preserving the text change depending on the selected option: 
![image](https://github.com/vocdoni/ui-scaffold/assets/6894329/c2d9a730-1c3c-48ad-b03d-762f951fee5b)
- [x] In the "Gitcoin passport" census, once selected add in the end of the description the following text "Read more about Gitcoin's Scorer & Thresholds here" with a link in the "here" that opens a new tab linking: "https://docs.passport.gitcoin.co/building-with-passport/major-concepts/scoring-thresholds"
- [x] The minimum possible GPS should be "1" instead of "0" (both for the numeric input & slider). 
- [x] The default GPS should be set to "20".
- [x] Correct the typo: `sybyl-attack` to `sybil-attack`.
- [x] In the "Token census", the description is as follows: "Configure the census by choosing a token from any available network. You can request a token here." but the "here" is not a link. Add the link to the "Add token form" (https://tally.so/r/mO46VY).
- [x] Change the text of the modal:
![image](https://github.com/vocdoni/ui-scaffold/assets/6894329/78e53d1b-9d9c-4644-8973-8fb029e81e03)
By removing: "Option not available in the free plan. To use it, you need to purchase the custom project plan." and replacing by: "XXXX is a Premium feature". (for instance: `Customization is a Premium Feature` or `SMS-based census is a Premium Feature`). 
Also, change the "Or you can directly schedule one quick call (30'') with our experts to help you create your desired voting project." to "Schedule a 30-minute call with our team to discuss your voting project. We'll guide you through the process and tailor solutions to meet your specific needs."
Same text for Spanish and Catalan.
- [x] Remove top margin space from the "Verify your vote" box and add some bottom margin.
![image](https://github.com/vocdoni/ui-scaffold/assets/6894329/20a9927f-f2c2-4cc3-a373-f7fec855671d)
- [x] Change "You can correct your vote only once." to "You can correct your vote once."
- [x] Add the "magic" emoji in the phrase: "Please wait while the anonymous magic happens  🪄."
- [x] When click in the "Request Custom Tokens", the form is open but also the option is selected in the dropdown. Prevent the dropdown to show this as an option: 
[...]

PR #662, opened 2024-04-25
Visual Improvements

selankon:
Fix #661 

From the list it fixes 

- [x] In the "Gitcoin passport" census, once selected add in the end of the description the following text "Read more about Gitcoin's Scorer & Thresholds here" with a link in the "here" that opens a new tab linking: "https://docs.passport.gitcoin.co/building-with-passport/major-concepts/scoring-thresholds"
- [x] The minimum possible GPS should be "1" instead of "0" (both for the numeric input & slider). 
- [x] The default GPS should be set to "20".
- [x] Correct the typo: `sybyl-attack` to `sybil-attack`.
- [x] In the "Token census", the description is as follows: "Configure the census by choosing a token from any available network. You can request a token here." but the "here" is not a link. Add the link to the "Add token form" (https://tally.so/r/mO46VY).
- [x] Remove top margin space from the "Verify your vote" box and add some bottom margin.
image
- [x] Change "You can correct your vote only once." to "You can correct your vote once."
- [x] Add the "magic" emoji in the phrase: "Please wait while the anonymous magic happens  🪄."
- [x] When click in the "Request Custom Tokens", the form is open but also the option is selected in the dropdown. Prevent the dropdown to show this as an option: 
image

PR #666, opened 2024-04-25
Implement union type custom radio button

selankon:
It fixes the first an issue of #661

> Improve OR/AND selector in the "Gitcoin Passport" census type. Change the current switch to the "custom radio button" Chakra component, while preserving the text change depending on the selected option:
> image

### DIFF
--- a/src/components/ProcessCreate/Census/Gitcoin/StampsUnionType.tsx
+++ b/src/components/ProcessCreate/Census/Gitcoin/StampsUnionType.tsx
@@ -58,10 +58,10 @@ export const StampsUnionType = () => {
 
   return (
     <Flex
-      gap={5}
-      flexDirection={{ base: 'column', md: 'row' }}
+      gap={2}
+      flexDirection={'column'}
       justifyContent={{ base: 'start', md: 'space-between' }}
-      alignItems={{ base: 'start', md: 'center' }}
+      alignItems={{ base: 'start', md: 'end' }}
       maxW={600}
     >
       <HStack>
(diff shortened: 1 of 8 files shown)


elboletaire (changes requested):
onvote theme should not be seen this way. You should be using a variant defined in both themes (like `primary`)

imatge

jpaulet:
Missing the "here" link with the actual link.

selankon:
> Missing the "here" link with the actual link.

@jpaulet the here link was fixed here https://github.com/vocdoni/ui-scaffold/pull/662 . This PR only fixes the AND/OR  selector

selankon:
> onvote theme should not be seen this way. You should be using a variant defined in both themes (like `primary`)

@elboletaire I just pushed a fix

src/components/ProcessCreate/Census/Gitcoin/StampsUnionType.tsx
```diff
@@ -1,55 +1,82 @@
...
+        {...checkbox}
+        cursor='pointer'
+        borderWidth='1px'
+        borderRadius='md'
+        boxShadow='md'
+        _checked={{
+          bg: 'primary.main',
+          color: 'white',
+          borderColor: 'primary.dark2',
+        }}
```
elboletaire, line 28: Sorry but this is still wrong. You should be using variables for everything, including the color set as `white`. Otherwise, if the `primary.main` color is too light, the text won't be legible. Ideally, you should use one of the variables used for buttons, since this looks like one.

elboletaire (changes requested):

src/components/ProcessCreate/Census/Gitcoin/StampsUnionType.tsx
selankon: After a bit talk I implemented this:  ced3008

elboletaire (approved):

PR #673, opened 2024-05-02
F/calculator

gerouvi:
TODO:

- [x] Add button link to stripe page
- [x] Fix UX button/link that leads to stripe to look like a button but also go to set the new uri
- [x] Add somehow the "Coming Soon" in the `Buy with Crypto` button
- [x] Explain with message that the minimum of 100 tokens when bying from stripe
   -  [x] Implement the logic of min amount  
- [x] If the user has already enough tokens for the voting process he wants to do show him a corrsponding message
- [x] separate faucet and calculator pages
- [x] Review which links go to the faucet and which to the caluclator

Election cost calculator

refs https://github.com/vocdoni/interoperability/issues/180


jpaulet:
Missing translations and we should select, by default, the "Single choice" voting, instead of none: 
image

src/i18n/locales/ca.json
```diff
@@ -30,6 +30,33 @@
...
+    "encrypted": "Encriptat",
+    "estimate_election_cost": "Estima el cost de la teva elecció",
+    "more_tokens": "Necessites més tokens dels indicats? <link>Contacta</link> amb el nostre equip de suport",
+    "num_elections": "Núm. d'eleccions",
+    "one_time": "Una sola vegada",
+    "package": "Paquet",
+    "packages": "Paquets",
+    "price_per_token": "Preu per token:",
+    "recurring": "Recurrent",
+    "recurring_description": "<p>Necessites una quantitat recurrent de tokens o tens la intenció de crear múltiples eleccions?<p/><p>Necessites votació digital segura pel teu software?</p><p>Oferim integracions atractives i opcions de serveis per a aquest tipus de necessitats</p><p>No dubtis i posa't en contacte amb nosaltres; el nostre equip d'experts t'ajudarà i t'elaborarà un pla específic per a tu!</p>",
```
elboletaire, line 52: Missing dot after `necessitats`.

elboletaire (changes requested):
This section is missing the final dot on sentences (in all languages). Also, add more padding in-between the sentences.
imatge

Also, we should add a new env var for this payment gateway (not sure if the stripe part can be toggled via env var @emmdim, but if it requires env vars it should be disabled when no env vars are set, and only allow crypto payments)

elboletaire (approved):

elboletaire (changes requested):
I approved it because looks good, but prettier complains about something wrong in these new files.

emmdim:
> I approved it because looks good, but prettier complains about something wrong in these new files.

Prettier issues fixed

vite.config.ts
```diff
@@ -27,6 +27,8 @@ const viteconfig = ({ mode }) => {
     defaultCensusSize = 5000
   }
 
+  let stripePublicKey = process.env.STRIPE_PUBLIC_KEY || ''
```
elboletaire, line 30: No need to to this here (do it directly in the JSON.stringify part), plus there's no need for a `let`. A `const` would be better.

jpaulet:
- [ ] When the "Calculate Cost" button is clicked, if the open tab is different than the "One-time", switch to this tab automatically.
- [ ] The "Min. 100 tokens" and "Coming soon" texts should be smaller.
- [ ] In the "If you have any question, check our FAQ or contact our support team" the "contact" word should be a link to "https://www.vocdoni.io/contact". 
- [ ] Change CA text from: "Si tens alguna pregunta, consulta les nostres Preguntes Freqüents o posa't en contacte amb el nostre equip de suport" to: "Si tens algun dubte, consulta les FAQ o posa't en contacte amb l'equip de suport". 
- [ ] Change ES text from: "Si tienes alguna pregunta, consulta nuestras preguntas frecuentes o contacta con nuestro equipo de soporte" to "Si tienes alguna duda, consulta las FAQ o contacta con el equipo de soporte".
- [ ] In the previous 2 tasks, add a anchor in the landing page to the FAQ section and link it with the word "FAQ" in all languages ( https://app.vocdoni.io/#faq )
- [ ] Move the 3 buttons used to switch between tabs outside the main container, in the top and as secondary buttons.
- [ ]  If the user is not logged-in and clicks the "Buy Tokens" button, show the login modal instead of the new blank page: 
image

- [ ]  Change the text: 

**How Vocdoni Works**

Vocdoni uses its own public voting blockchain (Vochain) to ensure maximum security, end-to-end verifiability, and a censorship-resistant voting process. To use this infrastructure, users need to expend tokens (a credit representation of the usage of the Vochain) to create the necessary election. Various parameters, such as census size and duration, determine the final voting cost.

Here you can simulate the cost of your voting process and purchase the needed tokens.

elboletaire:
IMHO the total should be shown always, otherwise selecting anything outside of "Once" will render the left side useless

imatge

jpaulet:
- [ ] The encrypted/overwrite flags are not working.
- [ ] If the cost is less than 100 tokens, do not disable the button. The purchase should still be possible, but for 100 tokens (even if only 45 are needed).
- [ ] In the "% Packages" tab, the different packages are not correctly sent to Stripe, and the amount always appears as "100".
- [ ] If an attempt is made to purchase more than 9999 tokens, the Stripe page crashes with the following error:

error 'session.New: {"status":400,"message":"If adjustable quantity is enabled, line_items[0][adjustable_quantity][maximum] (10000) must be greater than or equal to line_items[0][quantity] (16853). line_items[0][adjustable_quantity][maximum] can be set as high as 999999, but defaults to 99 if not set.","param":"line_items[0][quantity]","request_id":"req_8GVQ6c7WCTOIgY","request_log_url":"https://dashboard.stripe.com/test/logs/req_8GVQ6c7WCTOIgY?t=1721397251","type":"invalid_request_error"}'

- [ ] Remove/hide (for now) the text `A confirmation email will be sent to ...` until we actually send the email.
- [ ] If the user rejects to sign, the text should be: 

> You aborted the action. Without signing, you will not receive the purchased tokens. If this was an error, you can click on "Sign" to continue with the operation. If you want a refund, click on the "Refund" button.

- [ ] The "Get more" (tokens) button in the profile dropdown, should point to the calculator instead of the faucet.
- [ ] Add more vertical space between the header and the Stripe component.

src/components/Calculator/index.tsx
```diff
@@ -0,0 +1,674 @@
...
+import { Trans, useTranslation } from 'react-i18next'
+import { AiOutlinePercentage } from 'react-icons/ai'
+import { FaCcStripe, FaEthereum } from 'react-icons/fa'
+import { MdOutlineLoop } from 'react-icons/md'
+import { PiNumberSquareOneLight } from 'react-icons/pi'
+import { generatePath, Link as ReactRouterLink, useNavigate } from 'react-router-dom'
+import { useAccount } from 'wagmi'
+
+const TOKEN_PRICE = 0.15
+const MINIMUM_PURCHASE_TOKENS = 100
```
elboletaire, line 47: Define these in constants, following the constants naming we have there.

src/components/Calculator/index.tsx
```diff
@@ -0,0 +1,674 @@
...
+          </Flex>
+          <RadioGroup
+            value={String(totalTokens)}
+            onChange={(e) => {
+              if (e === '1000') setTotalTokens(1000)
+              else if (e === '15000') setTotalTokens(15000)
+              else if (e === '30000') setTotalTokens(30000)
+              else if (e === '50000') setTotalTokens(50000)
+              else if (e === '100000') setTotalTokens(100000)
+            }}
```
elboletaire, line 422: Wut? Why not simply...
```suggestion
            onChange={(e) => {
              setTotalTokens(Number(e))
            }}
```

src/components/Calculator/index.tsx
```diff
@@ -0,0 +1,674 @@
...
+const BuyBtns = ({ totalTokens }: { totalTokens: number }) => {
+  const { t } = useTranslation()
+  const navigate = useNavigate()
+  const { isOpen, onOpen, onClose } = useDisclosure()
+  const { account } = useClient()
+  const { isConnected } = useAccount()
+  const { openConnectModal } = useConnectModal()
+  const [hasConnected, setHasConnected] = useState(false)
+
+  const stripeEnabled = import.meta.env.STRIPE_PUBLIC_KEY.length > 0
```
elboletaire, line 612: Better have it in constants too.

src/components/Faucet/Stripe.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const { address } = useAccount()
+  const { client } = useClient()
+  const origin = window.location.origin
+
+  const fetchClientSecret = useCallback(async () => {
+    if (address) {
+      let uri = `${client.faucetService.url}/createCheckoutSession/${address}`
+      if (amount && !isNaN(parseInt(amount))) {
+        uri = `${uri}/${amount}`
+      }
```
elboletaire, line 36: Always specify the radix
```suggestion
      if (amount && !isNaN(parseInt(amount, 10))) {
        uri = `${uri}/${amount}`
      }
```

src/components/Faucet/Stripe.tsx
```diff
@@ -0,0 +1,227 @@
...
+      setStatus(data.status)
+      setCustomerEmail(data.customer_email)
+      setFaucetPackage(data.faucet_package)
+      setRecipient(data.recipient)
+    })
+
+    return () => {
+      abortController.abort()
+    }
+  }, [])
```
elboletaire, line 104: Kinda cumbersome to define a method that returns stuff and then use that return to set state values. Why not simply set state values during he function execution?

Also, why the infinite loop checking for a state value that won't be on the first run for sure??

src/components/Faucet/Stripe.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+  // claim the tokens if the package is not consumed
+  useEffect(() => {
+    if (
+      !faucetPackage ||
+      !recipient ||
+      !accountLoaded ||
+      packageConsumed ||
+      `0x${account?.address}` !== recipient.toLowerCase()
```
elboletaire, line 113: Use `ensure0x` rather than manually forcing the prefix

src/components/Faucet/Stripe.tsx
```diff
@@ -0,0 +1,227 @@
...
+    )
+  }
+
+  if (status === 'complete' && packageConsumed == true) {
+    return (
+      <Box as='section' id='success' className='site-wrapper' mt={10} textAlign='center'>
+        <Trans
+          i18nKey='claim.purchase_success'
+          components={{
+            customLink: <Link href='mailto:orders@vocdoni.org' />,
```
elboletaire, line 214: do no name it `customLink`. Name it `mailto` in this case.

src/elements/Stripe.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { useParams } from 'react-router-dom'
+import { CheckoutForm, CheckoutReturn } from '~components/Faucet/Stripe'
+
+export const StripeCheckout = () => {
+  const { amount }: { amount?: string } = useParams()
+
+  return (
+    <Box>
+      <CheckoutForm amount={amount} />
+    </Box>
```
elboletaire, line 11: These extra surrounding boxes look weird. Considering you're adding them on each view, and that these are routing elements, wouldn't make more sense to have such Box defined in the layout?

src/elements/Stripe.tsx
```diff
@@ -0,0 +1,22 @@
...
+  )
+}
+
+export const StripeReturn = () => {
+  const { sessionId }: { sessionId?: string } = useParams()
+  return (
+    <Box>
+      <CheckoutReturn sessionId={sessionId || ''} />
+    </Box>
```
elboletaire, line 20: Same, probably non-required wrapper.

src/themes/default/components/Home/index.tsx
```diff
@@ -10,23 +12,37 @@ import Solutions from './Solutions'
...
+  useEffect(() => {
+    if (hash) {
+      setTimeout(() => {
+        const element = document.getElementById(hash.substring(1))
+        if (element) {
+          element.scrollIntoView({ behavior: 'smooth' })
+        }
+      }, 0)
+    }
+  }, [hash])
```
elboletaire, line 27: Add a comment to the use effect to know its behavior.

src/themes/onvote/index.ts
```diff
@@ -34,6 +34,10 @@ export const theme = extendTheme(vtheme, {
       '.brand-theme': {
         fontFamily: 'pixeloidsans, monospace',
       },
+      '.calculator': {
+        border: '1px solid',
+        borderColor: 'calculator.border',
+      },
```
elboletaire, line 40: You should avoid using classes unless you don't have other options. In this case you should have defined a custom style object, like with any other custom component: https://v2.chakra-ui.com/docs/styled-system/component-style#styling-single-part-components

vite/features.ts
```diff
@@ -102,8 +103,6 @@ const features = () => {
     features._census[census] = true
   }
 
-  console.log('features:', features)
```
elboletaire, line 105: WHY? Get it back please. Change it to console.info if you feel better about it, but that console has to be there.

elboletaire (changes requested):

PR #678, opened 2024-05-03
Fix weight visualization

selankon:
It shows the votes of only the first question instead of the sum of votes for all questions.

If an election have more than one questions you expect to se number of votes emitted thinking that on the same ballot you answer all the questions. So looking only the first question results you can know how many votes are emitted

### DIFF
--- a/src/components/Process/Aside.tsx
+++ b/src/components/Process/Aside.tsx
@@ -38,7 +38,7 @@ const ProcessAside = () => {
   const showVotes = !election?.electionType.secretUntilTheEnd || election.status === ElectionStatus.RESULTS
 
   let votes = 0
-  if (election && showVotes) {
+  if (election && showVotes && election?.questions.length) {
     const decimals = (election.meta as any)?.token?.decimals || 0
     // It just has to check the first question to get the total of votes.
     const question = election?.questions[0]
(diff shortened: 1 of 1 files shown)


src/components/Process/Aside.tsx
```diff
@@ -37,23 +37,22 @@ const ProcessAside = () => {
...
   if (election && showVotes) {
     const decimals = (election.meta as any)?.token?.decimals || 0
-    const totalsAbstain = election?.questions.map((q) => ('numAbstains' in q ? Number(q.numAbstains) : 0))
-
-    totalWeight = election?.questions
-      .map((el, idx) => el.choices.reduce((acc, curr) => acc + Number(curr.results), totalsAbstain[idx]))
-      .map((votes: number) => results(votes, decimals))
-      .reduce((acc, curr) => acc + curr, 0)
+    // It just has to check the first question to get the total of votes.
+    const question = election?.questions[0]
```
elboletaire, line 44: Note all the previous operations could work if `questions` or `choices` were empty, because they were maps and reduces, but due to this change, and considering there's no `election?.questions.length` check, this will fail if the election has no questions defined.

elboletaire (changes requested):

src/components/Process/Aside.tsx
selankon: And this code is to count votes, so if there are no questions, where is supposed i can find votes?

elboletaire (approved):

PR #679, opened 2024-05-06
Labels removed except for the first question in the process

gerouvi:
Exclude the tag single and multiple choice for all questions except t he first one and more gap between questions and more gap between questions


elboletaire:
This has to be done in ui-components, since there's no reason to have this on each question considering how the ballot protocol works.

PR #685, opened 2024-05-14
Improve create election form states

selankon:
It create two new states on the form:

- isLoadingCost to know when is loading cost
- notEnoughBalance to disable the confirm button when not enough balance


elboletaire (commented):
I mean... it's not bad, but I feel like you overcomplicated a bit yourself.

Wouldn't all these changes be the same as just defining a const, say `isDisabled`  like...

```ts
const isDisabled = disabled || isLoadingPreview
```
and then use that `isDisabled` where `disabled` was used?

🤔

elboletaire (approved):

PR #696, opened 2024-05-16
fbg home and navbar

gerouvi:
Fixed issue, home background and navbar in local


elboletaire (changes requested):
Image is super heavy. Considering the type of image, it shouldn't be more than 50kb...

elboletaire (approved):

Issue #614, opened 2024-03-26
Implement dynamic page title


selankon:
After talk here https://github.com/vocdoni/ui-scaffold/issues/613 we decided to implement a dynamic page change. 

We found this hook which can do the job https://github.com/uidotdev/usehooks/blob/main/index.js#L265-L269

However, some questions raised to me during the implementation, i take note here ftr:

- [ ] Should a string be prepended on each title based on theme? Like `Onvote - Election title` or `Vocdoni - Election title`. And if yes, I am not sure of how to implement it yet. 
- [ ] Which pages should have custom title? Imo, election page, create election, organization page  and maybe not found page also. 
- [X] After test the hook, it don't go back to the previous title. A provider should be implemented with the default theme, and a system to go back to the default title. 

I started the implementation on https://github.com/vocdoni/ui-scaffold/tree/f/implement-document-title

PR #697, opened 2024-05-21
Change document title

selankon:
Fixes https://github.com/vocdoni/ui-scaffold/issues/614

It changes the document title on certain pages using the `useDocumentTitle` hook.

When the hook is unmounted it goes back to the default page title

### DIFF
--- a/src/elements/Process.tsx
+++ b/src/elements/Process.tsx
@@ -2,13 +2,13 @@ import { ConnectButton } from '@rainbow-me/rainbowkit'
 import { ElectionProvider, OrganizationProvider, useElection } from '@vocdoni/react-providers'
 import { PublishedElection } from '@vocdoni/sdk'
 import { useLoaderData } from 'react-router-dom'
-import { ProcessView } from '~components/Process/View'
+import { ProcessView as ProcessViewComponent } from '~components/Process/View'
 import { useDocumentTitle } from '~src/use-document-title'
 
-const SetProcess = () => {
+const ProcessView = () => {
   const { election } = useElection()
   useDocumentTitle(election?.title.default)
-  return <ProcessView />
+  return <ProcessViewComponent />
 }
 
 const Process = () => {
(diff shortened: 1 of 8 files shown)


elboletaire (approved):

src/elements/Process.tsx
```diff
@@ -1,16 +1,23 @@
...
+}
 
 const Process = () => {
   const election = useLoaderData() as PublishedElection
 
   return (
     <OrganizationProvider id={election.organizationId}>
       <ElectionProvider election={election} ConnectButton={ConnectButton} fetchCensus autoUpdate>
-        <ProcessView />
+        <SetProcess />
```
elboletaire, line 20: Terrible name 😅

selankon: True! 🙈 dbf3e00

elboletaire (changes requested):
I see a problem here... navigating through pages maintains one of the first titles set (?)

To reproduce:
- Go to your organization, you'll se the title change
- Go to the home page, the title remains with the org name

selankon:
FTR if we would like to centraliza all on an only one hook/component that just change the title, and dinamically check if is inside an election context or organization context to set the title, i did this approach which is not finished but can be used as inspiration if we need to do it on the future

```ts
export function useDocumentTitle(title?: string) {
  const defaultTitleRef = useRef(document.title)
  const electionCtx = useContext(ElectionContext)
  const orgCtx = useContext(OrganizationContext)

  useEffect(() => {
    const isOnvote = import.meta.env.theme === 'onvote'
    let prepend = 'Vocdoni'
    // Vocdoni - The voice of digital voting
    // Anonymous Gasless and Modular voting for Web3
    let title = 'The voice of digital voting'
    if (isOnvote) {
      title = defaultTitleRef.current
      prepend = 'OnVote'
    }
    if (electionCtx && electionCtx.election instanceof PublishedElection) {
      title = electionCtx.election?.title.default
    } else if (orgCtx && orgCtx.organization) {
      title = orgCtx.organization?.account.name.default ?? orgCtx.organization?.address ?? ''
    }

    document.title = `${prepend} - ${title}`
    return () => {
      document.title = defaultTitleRef.current
    }
  }, [electionCtx, orgCtx, title])
}
```

elboletaire:
> I see a problem here... navigating through pages maintains one of the first titles set (?)
> 
> To reproduce:
> 
>     * Go to your organization, you'll se the title change
> 
>     * Access a process, the title changes again
> 
>     * Go to the home page, the title remains with the election name (?)

I can still reproduce this 😕

elboletaire (approved):

PR #700, opened 2024-05-23
Delete CSP Election creation duplicated code

selankon:
This code seems to be duplicated, need to be tested

- [ ] For github elections
- [ ] For normal elections (should work as usual)

### DIFF
--- a/src/components/ProcessCreate/Steps/Confirm.tsx
+++ b/src/components/ProcessCreate/Steps/Confirm.tsx
@@ -101,12 +101,6 @@ export const Confirm = () => {
       }
       const election = Election.from(params)
 
-      let pid: string = ''
-      if (census instanceof CspCensus) {
-        pid = await client.electionService.nextElectionId(account!.address, election)
-        const createdCspElection: IElectionWithTokenResponse = await createElectionInCsp(pid, form.userList)
-      }
-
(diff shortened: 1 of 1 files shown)


src/components/ProcessCreate/Steps/Confirm.tsx
```diff
@@ -101,24 +101,16 @@ export const Confirm = () => {
...
           case ElectionCreationSteps.SIGN_TX:
           case ElectionCreationSteps.DONE:
             setStep(step.key as Steps)
             if (step.key === ElectionCreationSteps.DONE) {
-              if (pid !== step.electionId) {
-                pid = step.electionId
-                if (census instanceof CspCensus) {
-                  const createdCspElection: IElectionWithTokenResponse = await createElectionInCsp(pid, form.userList)
-                }
+              const pid = step.electionId
```
elboletaire, line 111: Careful. Doing it this way, the `setCreated` below won't know the actual pid assigned here, since you created a new scope.

elboletaire (changes requested):

elboletaire (approved):
This PR scares me a bit, NGL, but we're not actively using CSP right now, so I guess we'll find out when we get back to them.

PR #708, opened 2024-06-06
organization dashboard styles 

gerouvi:
Styles orgnization dashboard

### DIFF
--- a/src/components/Organization/Dashboard/ProcessCard.tsx
+++ b/src/components/Organization/Dashboard/ProcessCard.tsx
@@ -1,20 +1,83 @@
-import { Link } from '@chakra-ui/react'
-import { ElectionSchedule, ElectionTitle } from '@vocdoni/chakra-components'
+import { Badge, Box, HStack, Link, Text } from '@chakra-ui/react'
+import { ElectionStatusBadge, ElectionTitle } from '@vocdoni/chakra-components'
 import { useElection } from '@vocdoni/react-providers'
 import { ensure0x } from '@vocdoni/sdk'
+import { IoIosArrowRoundForward, IoIosCalendar } from 'react-icons/io'
+import { MdHowToVote } from 'react-icons/md'
 import { Link as RouterLink } from 'react-router-dom'
 import { ContentsBox } from '~components/Organization/Dashboard/Box'
 
 const ProcessCard = () => {
   const { election } = useElection()
 
+  const formatDate = (date: Date) => {
+    const options: Intl.DateTimeFormatOptions = {
+      month: 'short',
+      day: '2-digit',
+      year: 'numeric',
+      hour: '2-digit',
+      minute: '2-digit',
+      hour12: false,
+    }
+    const localeDateString = date.toLocaleString('en-US', options)
+
+    const [monthDay, year, time] = localeDateString.split(', ')
+    const [month, day] = monthDay.split(' ')
+
+    return `${month} ${day} - ${year} (${time})`
+  }
+
+  const calculateDaysUntil = (startDate: Date) => {
+    const now = new Date()
+    const timeDiff = startDate.getTime() - now.getTime()
+    const daysDiff = Math.ceil(timeDiff / (1000 * 3600 * 24))
+    return daysDiff
+  }
+
   if (!election) return null
 
   return (
-    <Link as={RouterLink} to={`/processes/${ensure0x(election.id)}`} w='full'>
+    <Link as={RouterLink} to={`/processes/${ensure0x(election.id)}`} w='full' textDecoration='none'>
       <ContentsBox w='full'>
-        <ElectionTitle fontSize='md' textAlign='left' />
-        <ElectionSchedule fontSize='xs' textAlign='left' />
+        <Box mb={3}>
+          {(election as any).status === 'UPCOMING' ? (
+            <Badge colorScheme='blue' px={2} py={1} borderRadius='lg' textTransform='lowercase' fontWeight='600'>
+              <Text as='span' textTransform='capitalize'>
+                Starting
+              </Text>{' '}
+              in {calculateDaysUntil((election as any).startDate)} days
+            </Badge>
+          ) : (
+            <ElectionStatusBadge />
+          )}
+        </Box>
+        <ElectionTitle fontSize='lg' textAlign='left' fontWeight='extrabold' />
+        <HStack mb={3}>
+          <MdHowToVote />
+          <Text fontWeight='600' fontSize='sm'>
+            {(election as any).voteCount} ({((election as any).voteCount / (election as any).census.size) * 100}%) votes
+            submited
+          </Text>
+        </HStack>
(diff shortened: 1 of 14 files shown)


src/components/Organization/Dashboard/ProcessCard.tsx
```diff
@@ -11,13 +16,76 @@ const ProcessCard = () => {
...
+          alignItems={{ base: 'start', lg: 'center' }}
+          justifyContent='space-between'
+        >
+          <ElectionStatusBadge fontSize='text-sm' mb={{ base: 3, lg: 0 }} order={{ lg: '2' }} />
+          <ElectionTitle fontSize='text' textAlign='left' fontWeight='extrabold' order={{ lg: '1' }} />
+        </Flex>
+        <HStack mb={3}>
+          <MdHowToVote />
+          <Text fontWeight='600' fontSize='text-sm' color='dashboard_card_text' className='dashborad_align_card_text'>
+            {(election as any).voteCount} ({((election as any).voteCount / (election as any).census.size) * 100}%) votes
```
elboletaire, line 32: Why are you using `as any` here? And also, round the result, i.e. `Math.round(value / 100) * 100`

src/components/Organization/Dashboard/ProcessCard.tsx
```diff
@@ -11,13 +16,76 @@ const ProcessCard = () => {
...
+            submited
+          </Text>
+        </HStack>
+        <HStack>
+          <IoIosCalendar />
+          <HStack>
+            <Text fontWeight='600' fontSize='text-sm' className='dashborad_align_card_text'>
+              <Text as='span' color='gray'>
+                {(election as any).startDate.getTime() < new Date().getTime() ? (
```
elboletaire, line 41: Same, you should not be using as any nowhere.

If the problem is because you cannot use `.startDate` on an InvalidElection, start by adding a `if (election instanceof InvalidElection) return null` at the beginning of the component.

src/components/Organization/Dashboard/ProcessCard.tsx
```diff
@@ -11,13 +16,76 @@ const ProcessCard = () => {
...
+                  <Trans i18nKey='dashboard.card.started'>Started</Trans>
+                ) : (
+                  <Trans i18nKey='dashboard.card.starts'>Starts</Trans>
+                )}
+                :
+              </Text>{' '}
+              <Text as='span' color='dashboard_card_text'>
+                {formatDate((election as any).startDate)}
```
elboletaire, line 49: Same

src/components/Organization/Dashboard/ProcessCard.tsx
```diff
@@ -11,13 +16,76 @@ const ProcessCard = () => {
...
+              <Text as='span' color='gray'>
+                {(election as any).endDate.getTime() < new Date().getTime() ? (
+                  <Trans i18nKey='dashboard.card.finished'>Finished</Trans>
+                ) : (
+                  <Trans i18nKey='dashboard.card.finish'>Finish</Trans>
+                )}
+                :
+              </Text>{' '}
+              <Text as='span' color='dashboard_card_text'>
+                {formatDate((election as any).endDate)}
```
elboletaire, line 64: And same

src/components/Organization/Dashboard/ProcessCard.tsx
```diff
@@ -11,13 +16,76 @@ const ProcessCard = () => {
   if (!election) return null
```
elboletaire, line 16: Do this and you won't need to use `as any`. Anyways, you should avoid using `as any` at all costs...  you could have used `as PublishedElection`, which would have been a bit better, but would still have the same issue we're having here (an invalid election will crash the app).

```suggestion
  if (!election || election instanceof InvalidElection) return null
```

src/components/Organization/Dashboard/ProcessCard.tsx
```diff
@@ -11,13 +16,76 @@ const ProcessCard = () => {
...
+
+  const locale = locales[i18n.language as keyof typeof locales] || enUS
 
+  const formattedDate = format(date, 'MMM dd, yyyy, HH:mm', { locale })
+
+  const [monthDay, year, time] = formattedDate.split(', ')
+  const [month, day] = monthDay.split(' ')
+
+  return `${month} ${day} - ${year} (${time})`
+}
```
elboletaire, line 90: Why?
1. We have methods to do this. We actually have a hook to format dates `const {format} = useDateFns()` will return the function you want, with all the locales stuff done.
2. This is not the place for a format date function, at all.

src/elements/Organization/Edit.tsx
```diff
@@ -1,10 +1,39 @@
...
+      <Flex
+        flexDirection={{ base: 'column', xl2: 'row' }}
+        justifyContent='space-between'
+        alignItems={{ base: 'start', xl2: 'center' }}
+        gap={2}
+      >
+        <Heading fontSize='heading-sm'>
+          <Trans i18nKey='organization.my_entity'>My entity</Trans>
+        </Heading>
+        <Button as={ReactRouterLink} to={`/organization/${address}`} variant='transparent' ml='auto'>
```
elboletaire, line 33: Either do not use a button here, or reduce its size, since it's making the header bigger in this section. I'd personally use a link, since you're actually using it as such, and it is not styled as a button.

src/themes/default/bg.tsx
```diff
@@ -1,18 +1,22 @@
...
+      bgImage={`url(${navbg})`}
+      bgSize='cover'
+      zIndex={-1}
+      filter='blur(20px)'
+      opacity={0.7}
+      backgroundSize='100%'
+    />
+    {children}
+  </>
 )
```
elboletaire, line 22:
```suggestion
export const NavBg = (props: BoxProps) => (
  <Box
    position='absolute'
    top={0}
    left={0}
    right={0}
    bottom={0}
    bgImage={`url(${navbg})`}
    bgSize='cover'
    zIndex={-1}
    filter='blur(20px)'
    opacity={0.7}
    backgroundSize='100%'
    {...props}
  />
)
```

elboletaire (changes requested):
Please review

src/i18n/locales/ca.json
```diff
@@ -670,16 +678,16 @@
...
-      "edit": "Editar organització",
       "read_more": "Expandir títol"
     },
-    "view_public_profile": "View public profile",
-    "voting_processes": "{{votes}} voting processes",
-    "votings": "Votings",
-    "votings_overview": "Votings overview"
+    "voting_processes_one": "{{ count }} procés de votació",
+    "voting_processes_many": "{{ count }} processos de votació",
+    "voting_processes_other": "{{ count }} processos de votació",
```
elboletaire, line 689: One of these two is redundant and unnecessary (not sure which one, but you can check by just removing one and then the other).

elboletaire (changes requested):

PR #713, opened 2024-06-17
footer styles

gerouvi:
Reduced footer size


elboletaire (changes requested):
I feel like this margin is incorrect. 
imatge

elboletaire (approved):

PR #734, opened 2024-07-08
removed votes count when the election is upcoming

### LINKED ISSUE #733: Aside votes
Remove votes from the process aside when the election hasn't started yet

gerouvi:
removed votes count when the election is upcoming.
 closes #733


elboletaire:
Applied a different solution 618f212dc9e70003278a2864f1b36146077aa573

PR #756, opened 2024-08-27
new views

gerouvi:
Team invite view


src/components/OrganizationSaas/Dashboard/Create.tsx
```diff
@@ -1,24 +1,167 @@
...
+        <Heading textAlign='center' mb='48px'>
+          <Trans>Try Vocdoni for free for 7 days</Trans>
+        </Heading>
+        <Box width='fit-content' mx='auto' mb='48px' fontWeight='500'>
+          <UnorderedList>
+            <li>Full access to basic features</li>
+            <li>Unlimited creation of voting processes</li>
+            <li>Multiple administrators</li>
+            <li>Up to 20 voters</li>
+            <li>Support during the trial period</li>
```
selankon, line 62: Should use `ListItem` instead of `li`? https://v2.chakra-ui.com/docs/components/list/usage#unordered-list

src/components/OrganizationSaas/Dashboard/Layout.tsx
```diff
@@ -11,25 +11,83 @@ import {
...
 import DarkModeToggle from '~src/themes/saas/components/DarkMode'
-import Wrapper from '~src/themes/saas/components/wrapper'
+import PricingCard from '~src/themes/saas/components/Saas/PricingCard'
+import Wrapper from '~src/themes/saas/components/Saas/Wapper'
 import useDarkMode from '~src/themes/saas/hooks/useDarkMode'
 import { Logo } from '~theme/icons'
 import OrganizationDashboardMenu from './Menu'
 import Settings from './Settings'
 
+const Cards = [
```
selankon, line 39: To add translation to this array  you will need to use the `useTranslation` hook, so the array should be declared inside a component.

Also, it should to be in lowecase (`const cards`). Another good practice is to define the inner object type (to ensure that all objects have the same attributes), something like:

```ts
type CardProps = {
popular: boolean
title: string
// etc
}
const cards: CardProps[] = [
//...
]
```

src/components/OrganizationSaas/Dashboard/Votings.tsx
```diff
@@ -16,16 +20,32 @@ const Votings = () => {
...
+        justifyContent='end'
+        sx={{
+          '& > div': {
+            display: 'flex',
+            flexDirection: 'column',
+            gap: '5px',
+          },
+          '& > div > div': {
+            ml: 'auto',
+          },
```
selankon, line 43: You could style the Pagination component itself maybe?? I don't think that is a good practice to do this style inheritance when we can just pass props to the Pagination component, but maybe @elboletaire have another opinion

src/components/OrganizationSaas/Dashboard/Create.tsx
```diff
@@ -1,24 +1,167 @@
...
+                base: '20px',
+                md: '44px',
+              }}
+            >
+              <Link fontWeight='500' href='https://simmmple.com/terms-of-service'>
+                Terms of Use
+              </Link>
+            </ListItem>
+            <ListItem>
+              <Link fontWeight='500' href='https://www.blog.simmmple.com/'>
```
selankon, line 156: Mailto and this two urls should point to the correct place better

src/components/OrganizationSaas/Dashboard/Layout.tsx
```diff
@@ -180,13 +253,57 @@ const OrganizationDashboardLayout: React.FC = () => {
...
+            difference. In the next billing period, starting on 'dd/mm/yy' you will pay for the new select plan.
+          </Trans>
+        </Text>
+        <Box>
+          <Text>
+            <Trans>Need some help?</Trans>
+          </Text>
+          <Button colorScheme='white'>
+            <Trans>Contact us</Trans>
+          </Button>
```
selankon, line 302: Is the contact us route already implemented? If is, navigate to it

src/themes/saas/components/Account/Create.tsx
```diff
@@ -0,0 +1,370 @@
+import {
```
selankon, line 1: not sure what is this component and why is here, does this do the same as signup? Or is another workflow?

selankon: And is correct to put this inside of themes folder?

src/themes/saas/components/modal.ts
```diff
@@ -0,0 +1,101 @@
...
+        '& > div:nth-of-type(2)': {
+          display: 'flex',
+          alignItems: 'center',
+          gap: '10px',
+
+          '& > p': {
+            whiteSpace: 'nowrap',
+          },
+        },
+        '& > div:last-of-type': {
```
selankon, line 81: In general, I don't think that do this is a good practice (styling the child using this kinds of css selectors). Maybe @elboletaire can suggest a better way to do this? Or we could just ignore it and go forward? 

Maybe creating a component and style the inner children it via props is a better approach?

selankon (changes requested):
I found that lots of `Trans` component have not the `i18nKey` prop, is this deliberated?

src/components/OrganizationSaas/Dashboard/Layout.tsx
gerouvi: Okay, Aulet told me that he would review the translations later. For now, I've left it so that it looks good visually

src/components/OrganizationSaas/Dashboard/Votings.tsx
elboletaire: Yeah, please, style components with their corresponding theme keys. Stop doing CSS in components with inheritance.

src/components/OrganizationSaas/Dashboard/Create.tsx
elboletaire: Also, if it's an external link, always use `isExternal`.

src/themes/saas/components/modal.ts
gerouvi: if I'm not mistaken, when creating a variant (of the modal in this case), we have to do it this way, since the parts of a modal (header, body, etc.) only extend so far, and to style the internal components, we need to access them in this way.

src/components/OrganizationSaas/Dashboard/Layout.tsx
gerouvi: It's not implemented yet, but I have set as={ReactRouterLink} so that it's ready. Thanks.

src/themes/saas/components/Account/Create.tsx
elboletaire: It is correct if the view requires a complete different layout for one theme compared to the other. But this specific view might need to be loaded when BACKEND_URL is defined, rather than given the theme saas.

gerouvi: It's another component https://github.com/vocdoni/interoperability/issues/201

src/themes/saas/components/modal.ts
elboletaire: Ideally, we should be using chakra-ui theming system (with multipart and normal). 

So rather than using `& > p` ideally we should define a section for that theme, and style it with the multi-part system, by name.

This will also help us in the future improve the theme and find issues, that otherwise would be hard to find due to the excess of CSS traversing.

elboletaire (changes requested):

src/components/OrganizationSaas/Dashboard/Layout.tsx
selankon: But you can just do like this

```
t('key.for.translation', { defaultValue: 'String you want to show' })
```

Then its easy to create the translations

src/themes/saas/components/modal.ts
selankon: To do not block this PR anymore I preffer to accept it and from now try to do this as you said, using the multipart component.

I am going to mark this as resolved

src/components/OrganizationSaas/Dashboard/Invite.tsx
```diff
@@ -0,0 +1,136 @@
...
+          <IoCloseSharp />
+        </Button>
+      </Flex>
+      <Text color={textColorSecondary}>
+        <Trans i18nKey='invite.subtitle'>Work together on projects</Trans>
+      </Text>
+      <HSeparator my='24px' />
+      <FormProvider {...methods}>
+        <Box as='form' onSubmit={methods.handleSubmit(onSubmit)}>
+          <Email label='Enter email' />
```
selankon, line 50: Label should be inside a translation also

src/elements/LayoutAuth.tsx
```diff
@@ -70,17 +78,17 @@ const LayoutAuth = () => {
...
               </ListItem>
               <ListItem>
-                <Link fontWeight='500' href='https://simmmple.com/terms-of-service'>
+                <Link fontWeight='500' href='/terms'>
                   {t('terms_of_use')}
                 </Link>
               </ListItem>
               <ListItem>
-                <Link fontWeight='500' href='https://www.blog.simmmple.com/'>
+                <Link fontWeight='500' href='https://blog.vocdoni.io/'>
```
selankon, line 91: missing `isExternal`

selankon (changes requested):

src/themes/saas/components/Account/Create.tsx
selankon: I will do this

src/components/OrganizationSaas/Dashboard/Layout.tsx
selankon: 18fe74e

src/components/OrganizationSaas/Dashboard/Invite.tsx
selankon: 18fe74e

src/elements/LayoutAuth.tsx
selankon: 18fe74e

src/themes/saas/components/Account/Create.tsx
```diff
@@ -0,0 +1,377 @@
...
+          </FormLabel>
+        </Flex>
+        <FormErrorMessage>{errors.terms?.message?.toString()}</FormErrorMessage>
+      </FormControl>
+      <Button form='process-create-form' type='submit' isLoading={create} mx='auto' mb='32px' w='80%'>
+        {t('organization.create_org')}
+      </Button>
+      <Text color={textColorSecondary} fontSize='sm'>
+        <Trans i18nKey='create_org.already_profile'>
+          {' '}
```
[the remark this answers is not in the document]
selankon: 18fe74e

src/components/AccountSaas/Create.tsx
jpaulet, line 1: IMO this view name should be more verbose, like "CreateOrg" or similar, in order to help finding it when needed.

src/components/OrganizationSaas/Dashboard/Create.tsx
```diff
@@ -1,24 +1,178 @@
...
+                <Image src={bloock} h={'30px'} />
+              </Flex>
+            </Grid>
+          </Box>
+          <Flex flexDirection='column' alignItems={{ base: 'center', lg: 'start' }} mb='100px' w='full'>
+            <Text fontSize={{ base: '28px', md: '48px', lg: '32px', xl: '48px' }} color='white'>
+              <Trans i18nKey='create_org.title'>The global voting platform</Trans>
+            </Text>
+            <Text fontSize='19px' color='white' textAlign={{ base: 'center', lg: 'start' }}>
+              <Trans i18nKey='create_org.subtitle'>Cut cost, Save Time: Secure, Private, and GDPR Compliant</Trans>
```
jpaulet, line 139: Typo: "costs"

src/components/AccountSaas/Create.tsx
selankon: In the actual scaffold this component has exactly this name:

`src/components/Account/Create.tsx`

I won't change it to follow same structure.

jpaulet (approved):

selankon (approved):

PR #758, opened 2024-08-28
Implement saas auth

selankon:


elboletaire (approved):
Understanding the "create organization" part is not directly related to this PR, so accepting...
imatge

PR #764, opened 2024-09-13
removed unnecessary variants

gerouvi:
Removed unnecessary variants in inputs and text area, and I moved the styles from the auth variant to baseStyles so that all inputs/text areas have the same styles. If we need different styles in the future, we will create a variant.

### DIFF
--- a/src/components/AccountSaas/Create.tsx
+++ b/src/components/AccountSaas/Create.tsx
@@ -133,7 +133,6 @@ export const AccountCreate = ({ children, ...props }: FlexProps) => {
             <Textarea
               {...register('description')}
               placeholder={t('form.account_create.description_placeholder').toString()}
-              variant='ss'
             />
           </FormControl>
         </Box>
--- a/src/components/AccountSaas/Create.tsx
+++ b/src/components/AccountSaas/Create.tsx
@@ -117,23 +117,20 @@ export const AccountCreate = ({ children, ...props }: FlexProps) => {
               {...register('name', { required })}
               placeholder={t('form.account_create.title_placeholder')}
             />
-            {!!errors.name && <FormErrorMessage>{errors.name?.message?.toString()}</FormErrorMessage>}
+            {!!errors.name && <FormErrorMessage>{errors.name?.message}</FormErrorMessage>}
           </FormControl>
           <FormControl isInvalid={!!errors.name} mb='24px'>
             <FormLabel ms='4px' fontSize='sm' fontWeight='500' color={textColor} mb='8px'>
               <Trans i18nKey='website'>Website</Trans>
             </FormLabel>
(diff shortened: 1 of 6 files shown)


jpaulet (approved):

src/components/AccountSaas/Create.tsx
```diff
@@ -134,13 +133,7 @@ export const AccountCreate = ({ children, ...props }: FlexProps) => {
...
               {...register('description')}
               placeholder={t('form.account_create.description_placeholder').toString()}
-              _focus={{
-                boxShadow: 'none',
-
-                _hover: {
-                  boxShadow: 'none',
-                },
-              }}
+              variant='ss'
```
selankon, line 136: Variant ss??

src/components/Auth/SignUp.tsx
```diff
@@ -78,15 +78,15 @@ const SignUp = () => {
...
+              <Input {...register('firstName', { required })} type='text' placeholder='First Name' />
               <FormErrorMessage>{errors.firstName?.message?.toString()}</FormErrorMessage>
             </FormControl>
             <FormControl isInvalid={!!errors.lastName} flexGrow={1} mb='24px'>
               <FormLabel display='flex' ms='4px' fontSize='sm' fontWeight='500' color={textColor} mb='8px'>
                 {t('signup_last_name')}
                 <Text color={textColorBrand}>*</Text>
               </FormLabel>
-              <Input {...register('lastName', { required })} variant='auth' type='text' placeholder='Last Name' />
+              <Input {...register('lastName', { required })} type='text' placeholder='Last Name' />
```
selankon, line 89: placeholders should be translated using `t` function

selankon (changes requested):

src/components/Auth/SignUp.tsx
```diff
@@ -78,15 +78,19 @@ const SignUp = () => {
                 {t('signup_first_name')}
                 <Text color={textColorBrand}>*</Text>
               </FormLabel>
-              <Input {...register('firstName', { required })} variant='auth' type='text' placeholder='First Name' />
+              <Input {...register('firstName', { required })} type='text' placeholder='First Name' />
```
selankon, line 81: This placeholder is still without translation :S

selankon (changes requested):

src/components/AccountSaas/Create.tsx
```diff
@@ -134,13 +133,6 @@ export const AccountCreate = ({ children, ...props }: FlexProps) => {
             <Textarea
               {...register('description')}
               placeholder={t('form.account_create.description_placeholder').toString()}
```
selankon, line 135: Unnecessary tostring

selankon (changes requested):

selankon (approved):

PR #765, opened 2024-09-13
Implement create account api

selankon:
### SDK

- Bumps vocdoni SDK and regenerates `yarn.lock`

### UI

- Improve `SelectCustom` component
- Implements create Saas account form logic
- Implements logout
- Implements mutation `useAccountCreate` to create an account to the Vocchain

### API

- Refactor API to a separated file
- Implement refresh token
- Re-instantiate RemoteSigner if berarer token is on LocalStorage
- Implements beared fetch to perform authenticated calls


src/components/Account/useAccountCreate.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const { exists } = useAccountHealthTools()
+
+  const create = async (values: CreateAccountParams) => {
+    let call = () =>
+      createAccount({
+        account: new Account(values),
+      })
+
+    if (exists) {
+      call = () => updateAccount(new Account(values))
```
emmdim, line 25: @selankon why is here the account updating implicit? Given that is a transaction that costs should it be done only when directly asked by the user?

src/components/Account/Create.tsx
```diff
@@ -33,13 +33,11 @@ import { ucfirst } from '~constants/strings'
 import { Check, Close } from '~theme/icons'
```
emmdim, line 33: @selankon `BasicAccountCreation` defined later in this file is not used anymore. Is it still necessary?

src/components/Account/Create.tsx
```diff
@@ -50,32 +48,22 @@ export const AccountCreate = ({ children, ...props }: FlexProps) => {
...
   })
+
   const {
-    createAccount,
-    updateAccount,
     errors: { account: error },
   } = useClient()
-  const { t } = useTranslation()
+
+  const { mutateAsync } = useAccountCreate()
```
emmdim, line 56: @selankon since the `createAccount` funcionality is now being done by the `useAccountCreate` hook shouldn't it expose as well the errors instead of importing them from `useClient`?

src/router/SignInScreen.tsx
```diff
@@ -1,9 +1,19 @@
...
 const SignInScreen = () => {
   const { openConnectModal } = useConnectModal()
+  const navigate = useNavigate()
+
+  // Redirect to the SAAS sign-in page
+  useEffect(() => {
+    if (!!import.meta.env.SAAS_URL) {
+      navigate('/signin')
+    }
+  }, [navigate])
```
emmdim, line 17: @selankon I do not comment on this because I undestand you will fix it in a future PR, right?

src/components/Auth/api.ts
```diff
@@ -0,0 +1,65 @@
+type MethodTypes = 'GET' | 'POST' | 'PUT' | 'DELETE'
```
emmdim, line 1: @selankon shoulnd't this file be in a different folder? Maybe the queries folder? or another utils folder?

src/components/Auth/useAuthProvider.ts
```diff
@@ -1,54 +1,34 @@
...
-    .catch((error: Error) => {
-      // todo(kon): implement error handling
-      throw error
-    })
-}
-
-const useLogin = (options: Omit<UseMutationOptions<LoginResponse, Error, ILoginParameters>, 'mutationFn'>) => {
-  return useMutation<LoginResponse, Error, ILoginParameters>({
-    mutationFn: (params: ILoginParameters) => api<LoginResponse>('auth/login', params, 'POST'),
+export const useLogin = (options: Omit<UseMutationOptions<LoginResponse, Error, ILoginParams>, 'mutationFn'>) => {
```
emmdim, line 21: @selankon Considering that there are going to be more queries needed in the future, I  think it makes sense to have the queries implmented in the query folder. What do you think?

src/components/Auth/useAuthProvider.ts
```diff
@@ -1,54 +1,34 @@
...
+  return useMutation<LoginResponse, Error, ILoginParams>({
+    mutationFn: (params: ILoginParams) => api<LoginResponse>(ApiEndpoints.LOGIN, { body: params, method: 'POST' }),
     ...options,
   })
 }
 
-const useRegister = (options: Omit<UseMutationOptions<LoginResponse, Error, IRegisterParameters>, 'mutationFn'>) => {
-  return useMutation<LoginResponse, Error, IRegisterParameters>({
-    mutationFn: (params: IRegisterParameters) => api<LoginResponse>('users', params, 'POST'),
+export const useRegister = (options: Omit<UseMutationOptions<LoginResponse, Error, IRegisterParams>, 'mutationFn'>) => {
```
emmdim, line 28: @selankon same as above

src/components/Auth/useAuthProvider.ts
```diff
@@ -58,49 +38,64 @@ enum LocalStorageKeys {
...
-  //       if (response.status !== 200) {
-  //         logout()
-  //       }
-  //     })
-  //     // network errors or other issues
-  //     .catch(() => {
-  //       logout()
-  //     })
-  // }, [])
+  const bearedFetch = useCallback(
```
emmdim, line 56: @selankon this seems like a generic function, as the one implemented in `api.ts`. 
Is it going to be used by future calls as well? If yes, it should be treated as `api.ts`, proably, outside the components folder

src/components/Auth/useAuthProvider.ts
```diff
@@ -1,54 +1,34 @@
-import { useMutation, UseMutationOptions } from '@tanstack/react-query'
 import { useClient } from '@vocdoni/react-providers'
```
emmdim, line 1: @selankon As a general recomendation, repeated in the comments later, what ever is not strictly login and logout, but amore generic reusable function should be probably moved in aseparate folder

emmdim (changes requested):

package.json
```diff
@@ -27,7 +27,7 @@
     "@tanstack/react-query": "^5.37.1",
     "@vocdoni/chakra-components": "^0.9.6",
     "@vocdoni/rainbowkit-wallets": "^0.2.2",
-    "@vocdoni/sdk": "~0.9.0",
+    "@vocdoni/sdk": "~0.9.1",
```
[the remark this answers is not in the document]
selankon: uups! thanks!

src/components/Account/useAccountCreate.tsx
selankon: I am not sure... This code is inherited from the old scaffold... I just moved the logic into a mutation to share it with the saas logic.  Take a look here https://github.com/vocdoni/ui-scaffold/blob/develop/src/components/Account/Create.tsx#L68-L79

Maybe this is an question for ElBoletaire

src/components/Account/Create.tsx
selankon: Well, I supose not. But this is code from old scaffold, take a look here... https://github.com/vocdoni/ui-scaffold/blob/develop/src/components/Account/Create.tsx#L127 should I delete it?

src/components/Account/Create.tsx
selankon: Mmm actually not, the errors came from the client because the functions called inside the `useAccountCreate` hook, `createAccount, updateAccount`, are processed on the react providers, and the errors are handled there. 

So, if the `createAccount, updateAccount` no error is thrown, and the errors are exposed on the `useClient` provider, afaik. 

So the mutation should never fail, the errors are handled inside the provider and updated the state with it.

Now talking about this, maybe the `useAccountCreate`  should return just the `create` function as a `useCallback` instead of wrap it in a mutation?

src/router/SignInScreen.tsx
selankon: Yes i am working on it on the https://github.com/vocdoni/ui-scaffold/pull/769 PR. But I have to understand first what exactly we want to do with this signin screen, there are some considerations to have. 

But, imo, lets talk this on the other PR when its ready.

src/components/Auth/useAuthProvider.ts
selankon: Same as on the

src/components/Auth/useAuthProvider.ts
selankon: Are you agree to take this changes on this PR https://github.com/vocdoni/ui-scaffold/pull/769 to merge this now?

package.json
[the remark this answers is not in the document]
selankon: 1a71deb

src/components/Auth/useAuthProvider.ts
selankon: 8060499

src/components/Account/Create.tsx
selankon: We decided to don't use a mutation, use a hook instead that expose the errors of the useClient

selankon: useAccountCreate

emmdim (approved):

src/components/Account/useAccountCreate.tsx
elboletaire: It's not an implicit call. That `call` method is called later when required

PR #766, opened 2024-09-16
F/custom input

gerouvi:
Refactored form components by removing all FormControl wrappers from the views. Implemented InputCustom and CheckboxCustom components to centralize field rendering and validation logic. These components handle Input and Checkbox fields respectively, simplifying view code and improving maintainability.


src/components/Auth/InputCustom.tsx
```diff
@@ -0,0 +1,94 @@
+export interface InputCustomProps {
```
selankon, line 1: Move this component into Layout folder

src/components/Auth/CheckboxCustom.tsx
```diff
@@ -0,0 +1,50 @@
...
+  label: ReactNode
+  required?: boolean
+  colorScheme?: string
+}
+
+const CustomCheckbox = ({
+  formValue,
+  label,
+  required = false,
+
```
selankon, line 18: uups this space

src/components/Auth/CheckboxCustom.tsx
```diff
@@ -0,0 +1,50 @@
+import { Box, Checkbox, Flex, FormControl, FormErrorMessage, FormLabel, Text } from '@chakra-ui/react'
```
selankon, line 1: move this to Layout folder

src/components/Auth/InputCustom.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export interface InputCustomProps {
+  formValue: string
+  label: string
+  placeholder?: string
+  type?: string
+  required?: boolean
+  validation?: any
+  messageError?: string
+}
```
selankon, line 35: You should add the chakra props for the FormLabel and Input so you will avoid to add type placeholder and others... Something like:

```ts
export interface InputCustomProps {
  formValue: string
  label: string
  required?: boolean
  validation?: any
  messageError?: string
  labelProps:  FormLabelProps
} & InputProps
```

(note that you don't need to declare `placeholder` or `type` properties because they are already defined on the chakra `InputProps`)

src/components/Auth/InputCustom.tsx
```diff
@@ -0,0 +1,94 @@
...
+  } = useFormContext()
+
+  const validationRules = {
+    ...validation,
+    ...(required ? { required: { value: true, message: t('form.error.field_is_required') } } : {}),
+  }
+
+  const errorMessage = errors[formValue]?.message?.toString() || ''
+
+  const inputType = type === 'password' && !show ? 'password' : type === 'password' ? 'text' : type
```
selankon, line 62: This is better to be something like

```ts
let inputType = type

if(type ===  'password' && show) {
     inputType = text
}
```

Much clear :) 

Try to do not use ternary operators unless is the only way.

src/components/Auth/CheckboxCustom.tsx
```diff
@@ -0,0 +1,50 @@
...
+  colorScheme = 'brandScheme',
+}: CustomCheckboxProps) => {
+  const { t } = useTranslation()
+  const { textColor } = useDarkMode()
+  const {
+    register,
+    formState: { errors },
+  } = useFormContext()
+
+  const validationRules = required ? { required: { value: true, message: t('form.error.field_is_required') } } : {}
```
selankon, line 28: Avoid to use ternary operators unless is absolutely needed.  Is much more readable if you do:

```
let validationRules = {}
if(required) {
    validationRules = { required: { value: true, message: t('form.error.field_is_required') } } 
}
```

Is much readable, try to use this method instead of the difficult to understand ternary opertaions

selankon (changes requested):

selankon (approved):

PR #767, opened 2024-09-18
Implement saas landing navbar

selankon:
It implements the landing page navbar for saas.

It only shows a button instead of the vocdoni app login modal an menus: if not authenticated it redirects to login page, else, to dashboard. 

Any functionality more.


jpaulet (approved):
For now I accept it, but we should add the "Logout" (with the "User Profile" icon).

emmdim (approved):

PR #768, opened 2024-09-18
Saas mail verification page

selankon:
Depends on https://github.com/vocdoni/saas-backend/pull/9/ and needs https://github.com/vocdoni/ui-scaffold/pull/765

It implements the email verification page. 

- On signup page, once signup is success it shows an "verification email sent" message with a "resend email button"
- On login page if the error of the backend is that the email is not verified, it shows the "verification email sent" message with a "resend email button"
- Implements the route `/account/verify?email=${email}&code=xxxxxxx`
- Performs an automatic verification attempt using a `useEffect`
- If success redirect to `/organization`
- Else show an error
- **Resend verification mail mutation is implemented but won't work until implemented on the backend**

In addition, it implements specific considerations for dev environments: 

- On signin page it shows a button to redirect to verify page without having verification code (because is not needed)
- It permits empty verification codes on verify page (on dev envs they are not neeeded)

An it fixes API function to:

- throw plain text errors

### DIFF
--- a/src/components/Auth/Verify.tsx
+++ b/src/components/Auth/Verify.tsx
@@ -1,0 +1,69 @@
+import { Box, Flex, FormControl, FormErrorMessage, Heading, Text } from '@chakra-ui/react'
+import { useTranslation } from 'react-i18next'
+import { useNavigate, useSearchParams } from 'react-router-dom'
+import useDarkMode from '~src/themes/saas/hooks/useDarkMode'
+import { useAuth } from '~components/Auth/useAuth'
+import { Loading } from '~src/router/SuspenseLoader'
+import { useEffect } from 'react'
+
+const Verify = () => {
+  const navigate = useNavigate()
+  const { t } = useTranslation()
+  const [searchParams] = useSearchParams()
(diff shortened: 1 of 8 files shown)


src/components/Auth/Verify.tsx
```diff
@@ -0,0 +1,124 @@
...
+      {import.meta.env.VOCDONI_ENVIRONMENT === 'dev' && (
+        <Button mt={4} as={ReactRouterLink} to={`/account/verify?email=${email}&code=`}>
+          Mail verification for dev envs
+        </Button>
+      )}
+    </Flex>
+  )
+}
+
+export default Verify
```
emmdim, line 124: @selankon I believe that this default export is unecessary. Can you please remove it and also update the import in the  router?

emmdim (approved):

src/components/Auth/Verify.tsx
selankon: We need default import to lazy load the page on router 

https://github.com/vocdoni/ui-scaffold/blob/53ee7cf570b909704e6352bc9f134c6e37fdcf10/src/router/Router.tsx#L37

PR #769, opened 2024-09-18
Improve auth state

selankon:
## Motivation

RemoteSigner class has a little difference with the normal Signer that is that `getAddress` function is asynchronous. 

This provoked some errors when accessing protected routes causing a blink show of Create Organization view on an "already logged in  with organization account".

To fix this I created an inner auth state to check if the RemoteSigner is loaded and has an address associated to it.

## Whats done

Add necessary authentication states on the provider to fix blinking page bugs when login in 

- Implements specific `SaasOrganizationProtectedRoute` to do not mix logics with the old scaffol `OrganizationProtectedRoute`
- Implements needed inner auth provider based on the signer state to properly redirect to create organization page. I used a Mutation to take advantage of the states that react query implements for this cases

### DIFF
--- a/src/components/Auth/useAuthProvider.ts
+++ b/src/components/Auth/useAuthProvider.ts
@@ -1,21 +1,52 @@
 import { useClient } from '@vocdoni/react-providers'
-import { RemoteSigner } from '@vocdoni/sdk'
+import { NoOrganizationsError, RemoteSigner } from '@vocdoni/sdk'
 import { useCallback, useEffect, useMemo, useState } from 'react'
 import { api, ApiEndpoints, ApiParams, UnauthorizedApiError } from '~components/Auth/api'
 import { LoginResponse, useLogin, useRegister, useVerifyMail } from '~components/Auth/authQueries'
+import { useMutation } from '@tanstack/react-query'
 
 enum LocalStorageKeys {
   AUTH_TOKEN = 'authToken',
 }
 
-export const useAuthProvider = () => {
-  const { signer, setSigner, clear } = useClient()
-  const [bearer, setBearer] = useState<string | null>(localStorage.getItem(LocalStorageKeys.AUTH_TOKEN))
+/**
+ * Mutation to set the RemoteSigner and check its address
+ * This hook is used as state to determine if an address is associated to the Signer to redirect
+ * to create organization page if not.
+ */
+const useSigner = () => {
+  const { setSigner } = useClient()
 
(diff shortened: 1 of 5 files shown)


src/router/Router.tsx
```diff
@@ -7,9 +7,10 @@ import Layout from '~elements/Layout'
 import LayoutAuth from '~elements/LayoutAuth'
 import LayoutProcessCreate from '~elements/LayoutProcessCreate'
 import { StripeCheckout, StripeReturn } from '~elements/Stripe'
-import OrganizationProtectedRoute from './OrganizationProtectedRoute'
+import { default as ScaffoldOrganizationProtectedRoute } from './OrganizationProtectedRoute'
 
 import { SuspenseLoader } from './SuspenseLoader'
+import SaasOrganizationProtectedRoute from '~src/router/SaasOrganizationProtectedRoute'
```
emmdim, line 13: @selankon The import should be from `'./SaasOrganizationProtectedRoute'`

src/components/Auth/useAuthProvider.ts
```diff
@@ -76,20 +99,19 @@ export const useAuthProvider = () => {
     clear()
   }, [])
```
emmdim, line 101: @selankon can you please elaborate on why you removed the refresh function?

emmdim (changes requested):

emmdim:
@selankon  I requested one minimal change and one explanation.
Other than that LGTM, if you want to merge it and I am not available go ahead

src/components/Auth/useAuthProvider.ts
selankon: Is not needed enymore. Refresh function was used before i know how that if I want to get the signer address i can just do

```
signer.address()
```

and all the reactive ui will respond to the new address. 

Refresh is not used anymore:

[src/router/Router.tsx](https://github.com/vocdoni/ui-scaffold/pull/769/files/6f0bce579f389f90fa85653a042d30b5ea8352c6#diff-12e851622e898e308434252678d097fb493488174f6fcf90bfff9fe09e7cdad6)

src/router/Router.tsx
selankon: ef732d4

jpaulet (approved):

selankon:
> @selankon I requested one minimal change and one explanation. Other than that LGTM, if you want to merge it and I am not available go ahead

This is done. I will dismiss the review because you are OoO today so we can merge

PR #773, opened 2024-09-23
Implement account protected pages

selankon:
## Motivation

Redirect to dashboard if you have an auth token and trying to access to signup, signin, forgot passoword o verify pages

### Changes

- It uses `Navigation` component instead of `navigate` inside a `useEffect` to redirect to signin page when trying to access on a `SaasOrganizationProtectedRoute` 
- It implements `SaasAuthProtectedRoute` to protect authentication routes if you have an auth token and redirect to dashboard so you can only logout. 
- **Breaking**: now signup and signin are inseide `/acocunt` path 
- Added a logout button on account create screen as suggestion by @jpaulet


jpaulet (approved):

emmdim (changes requested):
Please merge after #770

src/router/Router.tsx
```diff
@@ -169,41 +171,51 @@ export const RoutesProvider = () => {
...
-              {
-                path: 'account/verify',
-                element: (
-                  <SuspenseLoader>
-                    <Verify />
-                  </SuspenseLoader>
-                ),
+                children: [
+                  {
+                    path: 'signin',
```
emmdim, line 187: @selankon are you sure that signin and signup should under the `SaasAuthProtectedRoute`?  
Also maybe they signup part should be under `SaaSProtectedRoute`, meaning to have an address?
Btw the nameing begins to become quite vague

selankon: `SaasAuthProtectedRoute` in reality is a middleware that protect that if  you already logged in you cannot `signup`, `login`, `recoverMail`... withou a previous logout to avoid strange bugs. 

https://github.com/vocdoni/ui-scaffold/blob/026f62cfa70f50d7a8b82ed1674a38abeefe52c7/src/router/SaasAuthProtectedRoute.tsx#L12-L14

Agree that `SaasAuthProtectedRoute` is not the best name

emmdim (approved):

PR #778, opened 2024-09-30
Add mail verification code form

selankon:
It adds an input text to the mail verification page to copy paste the code if the user wants to do it (instead of follow the direct link sent on the email)


jpaulet:
> @selankon while technicaclly this PR is ok, from a UX perspective the fact that if you insert a wrong go then you are redirected to a new Error page and pressing the back button leads you to the inital page is problematic.
> 
> 2024-10-01-165746_875x479_scrot
> 
> I would suggest to show an error in the same page that asks for the verification code, and then allowing the user to inout the code again. cc @jpaulet

Yes, the error should go in the same page, with the instructions to re-check the code and input it again, or a new button to ask a new code, as commented by Manos.

selankon:
Totally agree. I also created a component `VerifyForm` to store the verify form logic sepparatelly

emmdim (approved):

PR #777, opened 2024-09-30
Implement edit profile screen

gerouvi:
A new view for profile editing.

It also does a refactor of the dashboard layout to allow scrolling while keeping the background always visible.

@selankon editions:

- Moved the `src/components/OrganizationSaas/Dashboard/EditProfile.tsx` into `src/components/AccountSaas/EditProfile.tsx`
- Created a `src/components/AccountSaas/Layout.tsx` to share inputs between forms 
- Same by shared types, created `src/components/AccountSaas/AccountTypes.ts`
- It implements the EditProfile page on `elements` folder

PR #779, opened 2024-10-02
Implement edit profile logic

selankon:
Depends on #777 

It implements logic and queries to perform a profile edition.

### Implements

- Creates a shared `QueryDataLayout` component to share the `isLoading` `isError` etc query logic to show loaders and error pages
- Implements `FormSubmitMessage` component to DRY
- Fixes `SelectCustom` component to show the label if only the value is set. On this way is easier to recover the information from the backend and use default value on the useForm hook. 
- Implements needed queries and mutations to edit profile on the SaaS


src/components/Layout/QueryDataLayout.tsx
```diff
@@ -0,0 +1,25 @@
...
+  if (isLoading) {
+    return <Loading minHeight={1} />
+  }
+  if (isError) {
+    return <ErrorComponent error={error} />
+  }
+  if (isEmpty) {
+    return null
+  }
+  return <>{children}</>
```
elboletaire, line 22: No need to use fragments here, simply return children.

src/components/Auth/api.ts
```diff
@@ -4,10 +4,11 @@ export enum ApiEndpoints {
...
   REFRESH = 'auth/refresh',
   REGISTER = 'users',
+  ORGANIZATIONS = 'organizations',
+  ORGANIZATION = 'organizations/{address}',
+  ORGANIZATION_MEMBERS = 'organizations/{address}/members',
   VERIFY = 'users/verify',
-  RESEND_VERIFICATION = 'users/verify/code',
-  ACCOUNT_CREATE = 'organizations',
-  TEAM_MEMBERS = 'organizations/{address}/members',
+  VERIFY_RESEND = 'users/verify/code',
```
elboletaire, line 11: Friendly reminder to change this from uppercase to PascalCase

elboletaire (changes requested):

elboletaire (approved):

PR #781, opened 2024-10-07
Create process form styles

gerouvi:
The styles from the SAAS process are applied, creating their different variations based on the default theme.


src/elements/LayoutProcessCreate.tsx
```diff
@@ -1,43 +1,67 @@
...
+              )}
+              <Button
+                as={ReactRouterLink}
+                onClick={(e) => (window.history.state.idx ? navigate(-1) : navigate('/'))}
+                colorScheme={isSaas && 'whiteAlpha'}
+                rightIcon={<CloseIcon boxSize={3} />}
+              >
+                {t('form.process_create.navbar.close_form_btn')}
+              </Button>
+              {isSaas && <DarkModeToggle />}
```
selankon, line 46: Theme not SAAS_URL

src/router/SaasProtectedRoutes.tsx
```diff
@@ -1,10 +1,10 @@
 import { Navigate, Outlet } from 'react-router-dom'
-import { useAccount } from 'wagmi'
 import { useAuth } from '~components/Auth/useAuth'
 
 const ProtectedRoutes = () => {
   const { signerAddress } = useAuth()
 
+  return <Outlet />
```
selankon, line 7: I think you forgot to delete this :)

src/themes/saas/index.ts
```diff
@@ -7,7 +7,7 @@ import { Badge } from './components/badge'
 import { Button } from './components/button'
 import { Input } from './components/input'
 import { Link } from './components/link'
-import { Form } from './components/Form'
+import { Form } from './components/form'
```
selankon, line 10: Some imports of this file are not well imported and rise lint issues.

selankon (changes requested):
Good job, this PR is huge! Try to find the way to do not rely on the `import.meta.env.SAAS_URL` to check if use styles for saas theme

src/themes/saas/components/checkbox.ts
```diff
@@ -1,26 +1,93 @@
 import { checkboxAnatomy } from '@chakra-ui/anatomy'
-import { createMultiStyleConfigHelpers } from '@chakra-ui/react'
+import { createMultiStyleConfigHelpers, defineStyle } from '@chakra-ui/react'
+import { mode } from '@chakra-ui/theme-tools'
 
-const { defineMultiStyleConfig } = createMultiStyleConfigHelpers(checkboxAnatomy.keys)
+const { definePartsStyle, defineMultiStyleConfig } = createMultiStyleConfigHelpers(checkboxAnatomy.keys)
 
+// Manteniendo la estructura original
```
elboletaire, line 7: Do not add comments in languages other than English, but also... this comment looks GPT-suspicious

elboletaire (changes requested):
Looks quite good, but there are a few things about the final result (which are hard to pinpoint directly in the code):
- The process creation had the option to create both multi-question and single-type questions, but with these changes, the single-type no longer works. Please try to restore it.
- I’m not a fan of adding buttons and features that aren’t immediately implemented. This applies to both the "approval" option added and all the new election options.
- Speaking of that new election options section that was added, I find it quite lacking. At the very least, it should have a title. And if the options haven’t been implemented yet, it might be a good idea to set all the unimplemented options as disabled until they’re properly developed (since I understand you added them for theming purposes and want to preview them).
- The use of color variables is quite wrong if the idea was to continue using themes... Remember, variable names for theme elements should be related to the section they belong to, rather than directly using a name like colorsBase.white.dark. But don’t worry about this—I’ll probably remove the theming system ASAP to reduce the burden it's causing on the project.

elboletaire (approved):

PR #826, opened 2024-10-29
saas voting

gerouvi:
Apply SaaS voting view with the same styles we had and dark mode


elboletaire (changes requested):
Disabled choices look weird both on dark and light themes. Other than that it LGTM

imatge
imatge

src/components/Process/View.tsx
```diff
@@ -106,8 +106,19 @@ export const ProcessView = () => {
...
+      position='relative'
+      width='full'
+      m='0 auto'
+      maxW='1920px'
+      px={{
+        base: '10px',
+        sm: '20px',
+        md: '80px',
+      }}
+      py={{ base: '0px', sm: '15px', md: '30px', md2: '30px', lg: '30px' }}
```
elboletaire, line 119: There's no need to repeat the same value after one definition. You should be able to safely remove md2 and lg from that list.

src/components/Layout/use-read-more.tsx
```diff
@@ -68,3 +79,37 @@ export const useReadMoreMarkdown = (
...
+      ? `${(containerMaxHeightPx * (tantPerCentGradient ? tantPerCentGradient : containerMaxHeightPx)) / 100}px`
+      : '0',
+    width: '100%',
+    bottom: 0,
+    background: isTruncated && readMore ? `linear-gradient(to bottom, ${fromDark} 0%, ${toDark} 100%)` : '',
+  },
+}}
+>
+{children}
+</Box> */
```
elboletaire, line 114: 😕

src/theme/index.ts
```diff
@@ -33,10 +35,50 @@ export const theme = extendTheme(vtheme, {
       ':root': {
         '--box-shadow': '0px 2px 4px lightgray',
         '--box-shadow-darker': '0px 2px 4px #808080b5',
+        '--read-more-from-light': 'rgba(245, 245, 247, 0)',
+        '--read-more-to-light': '#F5F5F7',
+        '--read-more-from-dark': 'rgba(26, 32, 44, 0)',
+        '--read-more-to-dark': '#1A202C',
```
elboletaire, line 41: There's no need to define custom variables this way. You can use the `colors.ts` to define the colors there, and then you can access them via `var(--chakra-colors-name-of-your-color)`.

elboletaire (changes requested):

elboletaire (approved):

PR #829, opened 2024-11-06
Resend verify code if out of date and trying to login

elboletaire:
Related to topics 2 & 4 from https://github.com/vocdoni/interoperability/issues/234

### DIFF
--- a/src/components/Auth/SignIn.tsx
+++ b/src/components/Auth/SignIn.tsx
@@ -59,7 +59,7 @@ const SignIn = () => {
   } = useAuth()
   const [verifyNeeded, setVerifyNeeded] = useState(false)
   const { mutateAsync: checkVerificationCodeStatus } = useVerificationCodeStatus()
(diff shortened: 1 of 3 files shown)


src/components/Auth/SignIn.tsx
```diff
@@ -33,19 +55,44 @@ const SignIn = () => {
   }, [])
 
   const {
-    login: { mutateAsync: login, isError, error, isPending },
+    login: { mutateAsync: login, isError, error },
   } = useAuth()
   const [verifyNeeded, setVerifyNeeded] = useState(false)
+  const { mutateAsync: checkVerificationCodeStatus } = useVerificationCodeStatus()
+  const resendVerificationCode = useResendVerificationCode()
```
emmdim, line 62: Why are the two above declarations different since they are being used in a similar manner later on?

emmdim (approved):
Funcionality seems correct, left a readibility question

PR #831, opened 2024-11-07
Password recovery flow

elboletaire:
Closes https://github.com/vocdoni/interoperability/issues/200

- Renamed some components for consistency
- Changed forgot route to `account/password` to reuse it for the reset (via `account/password/reset`)
- Edited the ForgotPasswordForm component to properly make expected API calls
- Created new ResetPasswordForm component to handle the password reset behavior
- Added optional email param to signin form to autofill it when trying to recover your password
- Created elements for both password reset and password forgot, moving the expected logic there
- Done the same for SignIn (creating elements/account/signin)
- InputBasic was not accepting additional props
- Added required routes

### DIFF
--- a/src/components/Auth/useAuthProvider.ts
+++ b/src/components/Auth/useAuthProvider.ts
@@ -62,7 +62,9 @@ export const useAuthProvider = () => {
         logout()
         throw new Error('No bearer token')
       }
-      headers.append('Authorization', `Bearer ${bearer}`)
+      if (bearer) {
+        headers.append('Authorization', `Bearer ${bearer}`)
+      }
       return api<T>(path, { headers, ...params }).catch((e) => {
         if (e instanceof UnauthorizedApiError) {
           return api<LoginResponse>(ApiEndpoints.Refresh, { headers, method: 'POST' })
--- a/src/components/Auth/useAuthProvider.ts
+++ b/src/components/Auth/useAuthProvider.ts
@@ -58,10 +58,6 @@ export const useAuthProvider = () => {
 
   const bearedFetch = useCallback(
     <T>(path: string, { headers = new Headers({}), ...params }: ApiParams = {}) => {
-      if (!bearer) {
-        logout()
-        throw new Error('No bearer token')
-      }
       if (bearer) {
         headers.append('Authorization', `Bearer ${bearer}`)
       }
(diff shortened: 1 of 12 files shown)


src/components/Auth/useAuthProvider.ts
```diff
@@ -62,7 +62,9 @@ export const useAuthProvider = () => {
         logout()
         throw new Error('No bearer token')
       }
-      headers.append('Authorization', `Bearer ${bearer}`)
+      if (bearer) {
```
emmdim, line 65: Why is this check needed while above `if (!bearer)` is also done?

src/elements/account/password/reset.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const code = searchParams.get('code') || ''
+  const email = searchParams.get('email') || ''
+
+  // Set layout title and subtitle
+  useEffect(() => {
+    setTitle(t('password_reset.title', { defaultValue: 'Password reset' }))
+    setSubTitle(
+      t('password_reset.subtitle', {
+        defaultValue:
+          "If your email is an existant account, you'll receive an email with a code to reset your account.",
```
emmdim, line 21: "If your mail corresponds to an existing* account, ..."

emmdim (approved):
Tested e2e and seems to work properly.
Left some minor comments

src/elements/account/password/index.tsx
```diff
@@ -0,0 +1,20 @@
+import { useEffect } from 'react'
```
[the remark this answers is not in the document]
elboletaire: In `elements` everything are pages following the defined routes (something I'm not very fond of, but at least it's standardized)... and I preferred having `account/password` so it was accessible (rather than defining different routes for each, or having `account/password` returning a 404).

src/components/Auth/useAuthProvider.ts
elboletaire: You are right... I was trying to add support so we could use `bearedFetch` for both public and private calls, but obviously it would not work, since it's throwing an error before...

But that logic doesn't make much sense anyway... if there's no token, why logout? We're only storing the bearer as session information, so if you get into that line, there's no need to logout...

I personally would remove that first check, and leave it open to be used for both public and private calls. It already handles login-out users in case of error, so right now... we would have an app crash for all these calls (if we logout the user, the next `bearedFetch` call would fail with the `No bearer token` error).

emmdim: LGTM

PR #835, opened 2024-11-12
Processes filtering logic

elboletaire:
- Moved react-query logic for the pagination and filtering to the route's `loader` method.
- Due to the above change, the `dashboard` routes were moved back to the hook, since having them externally would have increase significantly the amount of arguments passed to the function.
- Updated the sidebar links to point to these new filters.
- Removed old limit logic in ProcessesList, since we can now limit directly via the API.

What I don't like: the empty processes list layout...

### DIFF
--- a/src/queries/organization.ts
+++ b/src/queries/organization.ts
@@ -1,5 +1,6 @@
 import { useQuery } from '@tanstack/react-query'
 import { useClient } from '@vocdoni/react-providers'
+import { AccountData, FetchElectionsParameters, VocdoniSDKClient } from '@vocdoni/sdk'
 
 export const useLatestElections = (limit = 5) => {
   const { client, account } = useClient()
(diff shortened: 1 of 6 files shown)


src/queries/organization.ts
```diff
@@ -1,5 +1,6 @@
 import { useQuery } from '@tanstack/react-query'
 import { useClient } from '@vocdoni/react-providers'
+import { AccountData, FetchElectionsParameters, VocdoniSDKClient } from '@vocdoni/sdk'
 
 export const useLatestElections = (limit = 5) => {
```
selankon, line 5: This query appears to be unused

src/queries/organization.ts
```diff
@@ -13,12 +14,22 @@ export const useLatestElections = (limit = 5) => {
...
+  params: PaginatedElectionsParams
+) => ({
+  enabled: !!account?.address,
+  queryKey: ['organization', 'elections', account?.address, params],
+  queryFn: async () =>
+    client.fetchElections({
+      organizationId: account?.address,
+      page: params.page ? Number(params.page) - 1 : 0,
+      status: params.status?.toUpperCase() as FetchElectionsParameters['status'],
+    }),
```
selankon, line 34: Could be a good idea to type this return type using the same types as react query use:

```ts
): Pick<UseQueryOptions<ElectionListWithPagination>, 'enabled' | 'queryKey' | 'queryFn'>  =>({
```

Is not really needed, but just to ensure the types are correct. On this way, you ensure that the `queryClient.getQueryData` infers properly the return type also

src/router/routes/dashboard.tsx
```diff
@@ -19,76 +20,81 @@ const OrganizationTeam = lazy(() => import('~elements/dashboard/team'))
...
+          {
+            path: Routes.dashboard.processes,
+            element: (
+              <SuspenseLoader>
+                <DashboardProcesses />
+              </SuspenseLoader>
+            ),
+            loader: async ({ params }) => {
+              const query = paginatedElectionsQuery(account, client, params)
+              return queryClient.getQueryData(query.queryKey) ?? (await queryClient.fetchQuery(query))
```
selankon, line 84: This can be refactored to:

```ts
return queryClient.ensureQueryData(query)
```

Read more:

https://tkdodo.eu/blog/react-query-meets-react-router#getquerydata--fetchquery
https://tanstack.com/query/v4/docs/reference/QueryClient#queryclientensurequerydata

selankon (changes requested):
It works like a charm! I suggested some minor changes

src/queries/organization.ts
elboletaire: I'm not really sure why would I need to extra type stuff, when the type is properly inferred from the SDK
imatge

src/router/routes/dashboard.tsx
elboletaire: Thanks, changed 728455dd4b5d02fe1b1202534efc866562022ffe

selankon (approved):

PR #841, opened 2024-11-15
Subscription, plan & permissions

elboletaire:
What's done so far:
- Created the new "subscription" view (accessible via `/admin/subscription`)
- Recovered the menu link to this new page (under `Settings -> Subscription`)
- Display a table with the current subscription. It's been done as a table since it looks like it, but I'm not sure if it's useful considering there aren't plans for multiple subscriptions. ~~Responsive tables are also a problem with the `display: grid` we're using in the dashboard.~~ fixed in 9e2899c2ed0afa42654d4bcddda765b24d97b091
- Removed the temporary "open modal" button from the main menu
- Such modal can now be opened using the "view plans & pricing" button in the new page
- A new Subscription Provider/Context has been created, with a `permissions` method in order to check for permissions of the current plan, but it has not been applied yet anywhere
- The pricing modal has been minimally changed to allow setting a custom title, and moving its contents into an independent component
- Created upgrade tier and upgrade plan modals
- Created usePricingModals hook (with its respective context and provider)
- Integrate these modals into the invite team members flow, showing the modal for users trying to invite more memberships than allowed.
- Integrate them into the process create: voting types and features are now intercepted by these modals if an upgrade is required.
- Recovered the old form behavior, which fixed multiple issues like the approval voting not working
- Unified the tabs and checkboxes styles, both using the same `DetailedBox` component styles (since they are identical).
- Moved the `index.html` file back to the root path (its expected location).


jpaulet (approved):

src/components/Organization/Subscription.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export const Subscription = () => {
+  const { t } = useTranslation()
+  const { isOpen, onClose, onOpen } = useDisclosure()
+
+  return (
+    <>
+      <SubscriptionModal isOpen={isOpen} onClose={onClose} title={t('pricing.title')} />
+      <Button onClick={onOpen} alignSelf='end'>
+        View Plans & Pricing
```
jpaulet, line 27: Translation missing

src/components/Organization/Subscription.tsx
```diff
@@ -0,0 +1,91 @@
...
+          </Tr>
+        </Thead>
+        <Tbody>
+          <Tr>
+            <Td display='flex' alignItems='center' gap={3}>
+              <Avatar name={subscription.plan.name} size='sm' />
+              {subscription.plan.name} ({subscription.plan.organization.memberships} members)
+            </Td>
+            <Td>
+              <Tag>undefined</Tag>
```
jpaulet, line 71: Translation missing

src/components/TermsAndPrivacy/Privacy.tsx
```diff
@@ -1,20 +1,7 @@
...
-      sm: '20px',
-      md: '80px',
-    }}
-    mb={44}
-    mt={6}
-  >
+  <>
     <Box mb={5}>
       <Heading as='h1' size='xl' mb={3}>
         Privacy Policy
```
jpaulet, line 7: Translation missing

src/components/TermsAndPrivacy/Terms.tsx
```diff
@@ -1,20 +1,7 @@
...
-      sm: '20px',
-      md: '80px',
-    }}
-    mb={44}
-    mt={6}
-    mx='auto'
-  >
+  <>
     <Heading as='h1' size='xl'>
       Terms and Conditions for Vocdoni Association
```
jpaulet, line 6: Translation missing

src/components/TermsAndPrivacy/Terms.tsx
```diff
@@ -173,7 +160,7 @@ const Terms = () => (
       By continuing to use app.vocdoni.io or onvote.app, you confirm your acceptance of these terms and your agreement
```
jpaulet, line 160: Translation missing

yarn.lock
jpaulet, line 1: This field shouldn't be part of the git tracking...

jpaulet (changes requested):

src/components/Organization/Subscription.tsx
elboletaire: Actually, this should be the price. Was set as undefined because I had no way to know it then. I'll fix it.

PR #860, opened 2024-11-28
TabsCustom scaffold

gerouvi:
TabsCustom Component resolves the issue of making tabs responsive by converting them into a select, as they are not responsive by default in Chakra UI. There's an example on the home page to see how it works. Styling in the theme still needs to be done.


src/components/Layout/TabsCustom.tsx
```diff
@@ -0,0 +1,36 @@
+import { Box, Tabs } from '@chakra-ui/react'
+import { Select } from 'chakra-react-select'
+import { useState } from 'react'
+
+type TabOption = {
+  label: string
+  value: number
+}
+
+const TabsCustom = ({ children, tabProps, selectProps, selectBoxProps }: any) => {
```
elboletaire, line 11: Rename the component to `ResponsiveTabs`.

Also remove the `props` suffix from all props (not required since all the props are props for other components).

You can also change the `selectBox` to just `box`.

Finally, I think it'd be better if you used [the multipart component helper functions](https://v2.chakra-ui.com/docs/styled-system/component-style#styling-multipart-components), and styled it using the theme files. This would also allow us to add variants and reduce the amount of props used overall the entire project.

elboletaire (changes requested):

elboletaire (approved):

PR #874, opened 2024-12-18
removed unnecessary classes

gerouvi:
Removed unnecessary classes that we had from other themes, onvote...


elboletaire:
Home texts have a custom wrongly defined color

elboletaire (changes requested):
Home texts have a custom wrongly defined color for the light theme
imatge

elboletaire (approved):

PR #881, opened 2024-12-19
WIP rendering chakra components in lexical editor

elboletaire:


elboletaire:
Closing since I've not been able to properly do this... Quite complicated since the paragraph node can only be extended, and it does not allow decorator nodes... that + the fact that lexical uses DOM elements, and not react elements, doesn't help at all.

PR #887, opened 2024-12-19
Number 1 Questions step

### LINKED ISSUE #886: A number "1" appears before the "add option" button in process create questions step
Fixed the issue where a '1' appears after the options in the process create question step.

gerouvi:
Applied question number styles.

closes #886


elboletaire (changes requested):
Only multiquestion type elections should show this number. Right now the approval voting is still showing that number, and shouldn't be the case here.

PR #896, opened 2024-12-23
Dashboard menu

gerouvi:
Dashboard menu as Vocdoni App

### DIFF
--- a/src/components/Account/Menu.tsx
+++ b/src/components/Account/Menu.tsx
@@ -60,10 +60,10 @@ const AccountMenu: React.FC<BoxProps> = (props) => {
             <Trans i18nKey='logout'>Logout</Trans>
           </MenuItem>
           <MenuDivider />
-          <MenuItem fontSize={'xs'} as={RouterLink} to={Routes.terms} target='_blank'>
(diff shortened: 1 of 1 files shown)


src/components/Account/Menu.tsx
```diff
@@ -50,11 +50,22 @@ const AccountMenu: React.FC<BoxProps> = (props) => {
...
             <Icon as={LogOut01} />
             <Trans i18nKey='logout'>Logout</Trans>
           </MenuItem>
+          <MenuDivider />
+          <MenuItem fontSize={'xs'} as={RouterLink} to={Routes.terms} target='_blank'>
+            <Trans i18nKey='menu.terms'>Terms</Trans>
+          </MenuItem>
+          <MenuItem fontSize={'xs'} as={RouterLink} to={Routes.privacy} target='_blank'>
+            <Trans i18nKey='menu.privacy'>Privacy</Trans>
+          </MenuItem>
```
elboletaire, line 68: Should not be `target='_blank'`. Also, you should be using `isExternal` instead of target _blank, but here is not required at all.

elboletaire (changes requested):

elboletaire (approved):

PR #905, opened 2024-12-30
Dashboard menu styles

gerouvi:
Dashboard menu active styles and fixed issues color scheme gray

### DIFF
--- a/src/components/Dashboard/Menu/Item.tsx
+++ b/src/components/Dashboard/Menu/Item.tsx
@@ -25,12 +25,11 @@ export const DashboardMenuItem = ({
     onClick={hasChildren ? onToggle : undefined}
     isActive={hasChildren ? (isOpen ? true : false) : false} // Set active state
     justifyContent='start'
-    variant='transparent'
+    variant={isActive && !hasChildren ? 'underline' : 'transparent'}
     w='full'
     colorScheme='gray'
     leftIcon={icon ? <Icon as={icon} /> : undefined}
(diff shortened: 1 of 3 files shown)


src/components/Dashboard/Menu/Item.tsx
```diff
@@ -23,12 +23,15 @@ export const DashboardMenuItem = ({
     as={route ? ReactRouterLink : undefined}
     to={route ? generatePath(route) : undefined}
     onClick={hasChildren ? onToggle : undefined}
-    isActive={isActive} // Set active state
+    isActive={hasChildren ? (isOpen ? true : false) : false} // Set active state
```
elboletaire, line 26: Why...? The logic for being or not active was done in Options, why are you redoing the logic here again? If it wasn't working as expected, should have been fixed where it was being done.

elboletaire (changes requested):

src/components/Dashboard/Menu/Item.tsx
gerouvi: We discussed with @jpaulet applying different styles when an item is active, depending on whether it has children or not. In one case, a background color is used, and in the other, an underline. The problem is that I can’t filter whether it has children or not in the theme. That’s why, if it has children, I apply the styles in the theme, and if not, I apply the styles (underline) inline.

elboletaire: Ok, but it would have been better to have a variant for this, rather than hand-picking the styles. This way you only need to switch between the `transparent` and the `text` (or whatever) variant, rather than switching the `isActive` and adding inline styles.

elboletaire (approved):

PR #907, opened 2024-12-30
Tables

gerouvi:
Fixed issue with tables and striped variant in dashboard processes


elboletaire (changes requested):
Colors do not match the designs

imatge

designs:
imatge
imatge

Also, remember to recover the hover effect on rows.

elboletaire (changes requested):
Still with issues...
imatge
There's only a single border visible in the middle of the table, every row should have a border from what I can see in the original design. The colors aren't the same, the ones here are lighter (#fdfdfd) than the ones in the design (#fcfcfc). Also, I'm not sure we want the header in white, although that's one of the possibilities in the design.

Please also check this table where there's a border appearing unaligned to the other ones (under subscription section):
imatge

src/components/Organization/Dashboard/Votings.tsx
```diff
@@ -30,20 +31,22 @@ const VotingsList = ({ data, status }: VotingsProps) => {
...
-            <RoutedPagination pagination={pagination} />
-          </Flex>
-        </>
-      ) : !!status ? (
-        <NoResultsFiltering />
-      ) : (
-        <NoElections />
-      )}
-    </Flex>
+    <DashboardContents>
```
elboletaire, line 34: This should not be set in the component, but in the element instead, since it's what's giving it the dashboard style.

elboletaire (changes requested):

PR #930, opened 2025-01-09
Auth layout aside banner

gerouvi:
fixed style issues for the aside banner in the auth layout


elboletaire (changes requested):
This changes improve how these sections are seen on a 1080p screen, but I still see a couple of issues:
- The scrollbar is still visible
- When the content's still loading (probably only a SuspenseLoader is shown), the scroll increases quite a lot, then gets back to the current position when the content loaded.

imatge

elboletaire (approved):

PR #931, opened 2025-01-09
fixed styles recovery password

### LINKED ISSUE #910: Account Password recovery page improvement
In the recovery password page, we should improve the following:

- [ ] The "Send link ... " button should look as a primary button
- [ ] The "Sign In" text should be the same height as the other text
- [ ] The "Sign In" link should be the same as in other pages (login/register), with and look like a link.

image

gerouvi:
fixed styles recovery password, font sizes and button variant.

closes #910

### DIFF
--- a/src/components/Auth/PasswordForgotForm.tsx
+++ b/src/components/Auth/PasswordForgotForm.tsx
@@ -2,7 +2,7 @@ import { Button, Flex, Link, Text } from '@chakra-ui/react'
 import { useMutation } from '@tanstack/react-query'
 import { FormProvider, useForm } from 'react-hook-form'
 import { useTranslation } from 'react-i18next'
-import { NavLink, useNavigate } from 'react-router-dom'
+import { Link as RouterLink, useNavigate } from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


src/components/Auth/PasswordForgotForm.tsx
```diff
@@ -52,19 +52,17 @@ const PasswordForgotForm: React.FC = () => {
...
         <Text fontWeight='400' fontSize='sm'>
           {t('already_member')}
-          <NavLink to={Routes.auth.signIn}>
-            <Text as='span' color={'account.link'} ms={1} fontWeight='500'>
-              {t('signin')}
-            </Text>
-          </NavLink>
+          <Link as={NavLink} to={Routes.auth.signIn} ml={1} fontWeight={500}>
+            {t('signin')}
+          </Link>
```
elboletaire, line 65: IDK why `NavLink` was being used here, but it's a bit cumbersome, since this component adds styles for the currently active path, and it's not required at all for the login button.

Use the normal react-router-dom

elboletaire (approved):

PR #950, opened 2025-01-23
Fix (I hope) for the create org refresh issue

elboletaire:


jpaulet (approved):
Tested several times without being able to reproduce it. Then, we can (for now, and hopefully forever) mark as resolved.

PR #964, opened 2025-01-28
Dashboard aside buttons sticky and background

gerouvi:
Positioned the "New Voting" and "Logout" buttons in the dashboard aside with position: sticky at the bottom, and the dashboard background issue fixed.


elboletaire (changes requested):
Not a good solution and actually we should be fixing the entire menu, not just its footer

elboletaire (changes requested):
Much better, though there's a moment where it looks like this

imatge

I bet it's related to the addresses in the selector, but there should be a max width to avoid this

elboletaire (approved):

PR #971, opened 2025-01-31
dashboard aside button gap

gerouvi:
Fixed issue dashboard aside gap when is active


elboletaire (changes requested):
@gerouvi you requested a new review but the buggy behavior is still there...

https://github.com/user-attachments/assets/2dca15a7-0a8b-4c78-92bf-cb861fb50b1c

Note: the video looks accelerated compared to what actually happens. It takes around one second from one change to the other, which makes this too much visible. Also, I still see a lot of gap compared to the example you did show me via private chat.

src/components/Dashboard/Menu/index.tsx
```diff
@@ -23,7 +23,8 @@ const DashboardMenu = ({ isOpen, onClose }: { isOpen: boolean; onClose: () => vo
       position={'fixed'}
       height={'100vh'}
       gridAutoColumns={'auto 1fr auto'}
-      maxW={'220px'}
+      maxW={'250px'}
+      minW={'250px'}
```
elboletaire, line 27: Simply use width instead

elboletaire (changes requested):

PR #977, opened 2025-02-04
Subscription no organization

### LINKED ISSUE #955: Add "No org" component in Subscription page
When the user don't have an organization, the subscription plan appears empty. Add the "No org" component, as in the others pages.

Also, the "View Plans & Pricing" show point to /plans (or at least open the modal, now the button does noting).

Image

gerouvi:
Render no organization component and view plans button links to plans view.

closes #955


elboletaire (changes requested):
This is not the way to do this, there's a protected route (`OrganizationProtectedRoute`) that should be used instead.

PR #991, opened 2025-02-12
fixed terms and privacy issues

gerouvi:
Reduced the size of titles and subtitles in the Terms and Privacy view, removed the padding x and fixed the links in the footer of Process Create.


elboletaire (approved):
Accepted, though in these cases you should consider creating a custom component (no need to create it in a different file, simply create a silly styled component in the same file)

PR #992, opened 2025-02-13
PlansLayout padding 

gerouvi:
Removed padding X from the PlansLayout component.


elboletaire (changes requested):
What's the sense on removing this padding and adding it again on each view using this layout??
imatge

PR #995, opened 2025-02-13
Navbar

gerouvi:
Fixed navbar issue using Dashboard component


elboletaire (changes requested):
Please use a defined route for the `useMatch()` checks (`Routes.processes.view` probably)

elboletaire (approved):

PR #1008, opened 2025-02-25
filter processes list

### LINKED ISSUE #1005: Delete button set when no processes available
Don't show the button set of "All/Active/Finished" in the page when no voting processes are shown.

Image

gerouvi:
Removed filter buttons when there are no elections.

closes #1005


elboletaire (changes requested):
Not a valid solution, since right now the filters are being hidden (and not in a good way) when you change to a filter with zero results

imatge

elboletaire (changes requested):
Rather than storing to state something (which will work, but causes some side-effects, like if you access via URL, you won't see the filters even having processes in other sections), use the `organization.electionIndex` (via `useOrganization` hook) attribute to determine if there are (or not) elections.

PR #1009, opened 2025-02-25
Pagination buttons styles

### LINKED ISSUE #1003: Improve pagination UI
The pagination shouldn't have the green background, only the border and all in black (text and color). 

Image

gerouvi:
Background transparent, color and border color black in light mode and white in dark mode.

 closes #1003

### DIFF
--- a/src/theme/components/pagination.ts
+++ b/src/theme/components/pagination.ts
@@ -1,4 +1,4 @@
-import { createMultiStyleConfigHelpers } from '@chakra-ui/react'
+import { createMultiStyleConfigHelpers, defineStyle, defineStyleConfig } from '@chakra-ui/react'
 import { PaginationAnatomy } from '@vocdoni/chakra-components'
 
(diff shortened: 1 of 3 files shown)


src/theme/components/pagination.ts
```diff
@@ -15,6 +15,21 @@ export const Pagination = defineVoteWeightTipStyle({
...
+
+        _hover: {
+          color: 'black',
+        },
+
+        _dark: {
+          color: 'white',
+          borderColor: 'white',
+        },
+      },
```
elboletaire, line 32: Why are you doing CSS traversing? You have a `PageButton` component you can style on its own.

Also, I don't see how the current page is being differentiated from others (?)

elboletaire (changes requested):

elboletaire (changes requested):
There's no way to see which is the active/current page

Issue #986, opened 2025-02-11
Cookies Banner Component


jpaulet:
As a visitor to our voting SaaS landing page, I want to see a cookie banner that matches the look & feel of the current page, so that I can be informed about our cookie policy and make a choice about my data.

**Acceptance Criteria:**

- [ ] A cookie banner appears at the bottom of the page when a user visits for the first time.
- [ ] The banner includes a generic text about cookies and links to the Privacy Policy and Terms & Conditions pages.
- [ ] Users can choose between three options:
        - Accept (stores consent and dismisses the banner).
        - Cancel (dismisses the banner without accepting).
- [ ] The choice is stored in localStorage to prevent the banner from appearing again.
- [ ] Consideration: Optionally store user consent in the database for compliance.
- [ ] The banner’s design aligns with the current Chakra UI theme.

Try to search a custom pre-made Chakra component or create it. Check 

` npm install react-cookie-consent`

Import the CookieConsent component and use it in your application:

```
import CookieConsent from "react-cookie-consent";

function App() {
  return (
    <div>
      {/* Your existing components */}
      <CookieConsent
        location="bottom"
        buttonText="Accept"
        declineButtonText="Cancel"
        enableDeclineButton
        onAccept={() => {
          // Handle acceptance logic
        }}
        onDecline={() => {
          // Handle decline logic
        }}
        style={{ background: "#2B373B" }}
        buttonStyle={{ color: "#4e503b", fontSize: "13px" }}
        declineButtonStyle={{ color: "#fff", background: "#c00", fontSize: "13px" }}
      >
        This website uses cookies to enhance the user experience.{" "}
        <a href="/privacy-policy" style={{ color: "#4e9c81" }}>
          Privacy Policy
        </a>{" "}
        and{" "}
        <a href="/terms-and-conditions" style={{ color: "#4e9c81" }}>
          Terms and Conditions
        </a>.
      </CookieConsent>
    </div>
  );
}

export default App;

```

And customize it, to ensure the cookie consent bar matches your application's look and feel, you can customize its styles. Since you're using Chakra UI, you can leverage its theming capabilities to maintain consistency. For instance, you can use Chakra's useTheme hook to access theme values:

```
import { useTheme } from "@chakra-ui/react";
import CookieConsent from "react-cookie-consent";

function App() {
  const theme = useTheme();

  return (
    <div>
      {/* Your existing components */}
      <CookieConsent
        location="bottom"
        buttonText="Accept"
[...]

PR #1011, opened 2025-03-01
feat(ui): add cookie consent banner

rahulsingh3526:
refs #986


elboletaire (changes requested):
Overall looks ok, but needs adding i18n. Use the `Trans` component and the `useTranslation` hook in order to add translatable strings.

After converting all the translatable strings to use the component (or the hook), run `yarn translations` to extract them all and commit the locales files.

There's no need to actually translate them to spanish and catalan, we'll do that 😊

PR #1012, opened 2025-03-02
Sidebar dashboard

### LINKED ISSUE #1004: New Sidebar
Adapt the sidebar with the following changes:

- [ ] Vocdoni logo to only the icon, centered and smaller.
- [ ] Create new "Workspace" component, with the different orgs on it and the "Create new org" inside, instead of the "+" icon.
- [ ] Organize the sidebar menu items and add new sections/items.
- [ ] Move the "Profile" inside the User Profile menu.
- [ ] Add the "Subscription component".
- [ ] Only add the "Upgrade" if the organization is in the free plan.
- [ ] Move the "Profile" into the bottom part of the sidebar and improve its UI.

Use the components in the side of the dashboard to adapt the wire-frame, the wire-frame is only for visual reference of the content but the final look and feel should come more from the UntitledUI components:

Image

gerouvi:
Restyle dashboard sidebar as required.

closes #1004


elboletaire:
closing since the changes have nothing to do with this in the end

PR #1029, opened 2025-04-02
Refactor dashboard layout

### LINKED ISSUE #1027: Adapt the dashboard layout to latest specs
- [ ] Layout should have a collapsible menu with the button always visible nearby it
- [ ] Menu has a slight gray background, contents are all white with no other separation than the border in the menu
- [ ] There's no delimiter for the main contents, but it has a max-width and is centered in the middle
- [ ] The top bar now has a breadcrumb (configurable via template vars), the section title disappears from there and also we visibly separate this top bar with a border
- [ ] Make the create process use the common dashboard layout, and remove its currently defined custom layout.

So the idea is to have something visually similar to what @ferranrego presented as mockups, but at the same time lose the minimum time changing styles, since we already implemented most of the untitled ui styles.

Here's a screenshot from an example dashboard design in untitled UI, we're going into this direction but switching the backgrounds (the sidebar has a grayed bg, the contents don't):

Image

Since the contents will have a white background and the sections are white too, everything must have border and a **slight** shadow.

But this is our reference, contently and structurally speaking: https://v0-modern-voting-dashboard.vercel.app/dashboard

gerouvi:
Dashboard layout changes:

- Refactor of the entire layout structure
- Removed content styles along with the background
- Toggle for the sidebar menu on desktop
- Breadcrumb
- Modified theme files such as buttons, headings, links, colors, radius, and spacing to adapt them to the requirements
- Removed the LayoutProcessCreate component

closes #1027


src/components/Dashboard/ProcessView.tsx
```diff
@@ -56,9 +56,11 @@ export const ProcessView = () => {
   const { t } = useTranslation()
 
   useEffect(() => {
-    setTitle(t('vote_details', { defaultValue: 'Vote details' }))
-    setBack(generatePath(Routes.dashboard.processes))
-  }, [setTitle, setBack, t])
+    setBreadcrumb([
+      { title: t('organization.dashboard'), route: Routes.dashboard.base },
```
elboletaire, line 60: Remove this extra breadcrumb

src/elements/LayoutDashboard.tsx
```diff
@@ -1,79 +1,129 @@
...
+                        {t('organization.dashboard')}
+                      </Link>
+                    </ListItem>
+                    <ListItem display={'flex'} justifyContent={'center'} alignItems={'center'}>
+                      <Icon as={ChevronRight} />
+                    </ListItem>
+                  </>
+                ) : (
+                  <Text>{t('organization.dashboard')}</Text>
+                )}
```
elboletaire, line 90: It doesn't feel ok that the layout is different when is linked compared when it's not. And Ordered (or Unordered) List must have ListItem as children, so this is not correct. The `Text` should be (or be wrapped with) a ListItem too.

elboletaire (changes requested):

elboletaire (approved):

PR #1039, opened 2025-04-29
refactor: auth layout and views

### LINKED ISSUE #1020: Apply the new login/signup page layout and functionality based on the updated Figma design
**Description**
Apply the new login/signup page layout and functionality based on the updated Figma design.

**Criteria**
* Must follow the Figma design provided.
* Must include a carousel displaying 4 different testimonials (See copies in the comments)
* Must simplify the login/signup flow:
    * When a user enters their email, the system checks if it’s in the database.
    * If it exists, prompt for password.
    * If it doesn’t, guide the user to create a new account.

gerouvi:
closes #1020


src/components/Auth/SignUp.tsx
```diff
@@ -84,75 +86,116 @@ const SignUp = ({ invite }: SignupProps) => {
...
+              validation={{
+                required: t('form.error.field_is_required'),
+                validate: (value) => value === password || t('passwords_do_not_match'),
+              }}
+            />
+            <FormControl as='fieldset' isInvalid={!!errors?.terms}>
+              <Checkbox
+                {...register('terms', { required: t('cc.validation.required') })}
+                color='auth.checkboxes_text'
+                isRequired
```
elboletaire, line 133: Remove the `isRequired`, since right now this is the only error you'll see initially until marked. Removing the `isRequired` (but keeping the validation) gives a better UX.

elboletaire (changes requested):
Overall looks good, other than the reviewed comments, try to fix this weird behavior that extends the layout to the entire height of the window while loading:


https://github.com/user-attachments/assets/de844867-200a-4c07-8e3e-d68320defa27

elboletaire (changes requested):
There's an error on how the data is captured in the form. Even though the mockups show the name as a single field in the signup, it's actually a two field part: name and surname. 

Also, I don't think there's a need to ask for a repeated password having the input fields with the toggle button to show them, but I need @ferranrego to confirm this, since it's a specific change he added in his mockups.

ferranrego:
Yes, let's remove the 2nd password field, it simplifies the form and we can always reintroduce it later if we think different in the future.

As for the name field, even though I didn’t reflect it in the mockups, it should be split into two fields (name and surname) everywhere in the UI, as @elboletaire pointed out. I’ll update the mockups accordingly.

elboletaire (approved):

elboletaire:
Yeah and also remove these paddings here: 

imatge

ferranrego:
When a user enters an incorrect 2FA code, the UI currently displays the error message:
“authentication required”

It should be updated to:
“The code you entered is incorrect. Please try again”

image

ferranrego:
Overall everything looks good! I made a review and I would ask for some small changes

**Create user account step:**
- [ ] Update the text “Don’t have an account? Create an account” to “Don’t have an account? Sign up”
- [ ] Change the labels “Name” and “Surname” to “First name” and “Last name”
- [ ] Ensure the checkboxes use the same border style as the other form fields. Example:
image

**Email verification step:**
- [ ] The email verification step should look like this:
image
- [ ] In the email verification step, if the user introduces the correct code, it should jump into the next step automatically, even if the user doesn't click on "Verify"

**Organization creation step:**
- [ ] The organization creation form should look like the image and [this video](https://screen.studio/share/anM8dqy5) (including copies)
image
- [ ] Remove the X that appears in the Country field when a country is selected.
image

Issue #1018, opened 2025-03-25
Apply the new visual theme configuration across the app using Chakra


ferranrego:
- [ ] Map the defined UntitledUI styles* (colors, shadows, radii, border, fonts, textStyles,...) on the project theme and components.
- [ ] Ensure all icons are sourced from Untitled UI icon's library, unless a required icon is not available.

PR #1053, opened 2025-05-15
refactor: theme

elboletaire:
refs #1018


elboletaire:
I'm accepting this now to have dev in a better state

PR #1060, opened 2025-05-23
feat(team): Adding team actions

AlexanderArce:
Adding the buttons for changing role, removing member and resend/cancel invitation

### DIFF
--- a/src/components/Organization/Team.tsx
+++ b/src/components/Organization/Team.tsx
@@ -22,15 +22,19 @@ import {
   PopoverFooter,
   PopoverTrigger,
   Progress,
+  Radio,
   RadioGroup,
+  RadioProps,
   Stack,
   Text,
   useDisclosure,
+  useRadioGroupContext,
   useToast,
 } from '@chakra-ui/react'
 import { useMutation, useQuery, useQueryClient, UseQueryOptions } from '@tanstack/react-query'
 import { enforceHexPrefix, useClient } from '@vocdoni/react-providers'
 import { formatDistanceToNow } from 'date-fns'
+import { ReactNode } from 'react'
 import { Controller, FormProvider, useForm, useFormContext } from 'react-hook-form'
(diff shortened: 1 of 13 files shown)


AlexanderArce:
### Things to Review

- **Missing actions for "Cancel Invitation" and "Resend Invitation"**:  
  Currently, there are no API endpoints available for these actions. Attempting to use the `Send Invitation` call with the same user as a workaround for `Resend Invitation` didn’t work—most likely due to **membership plan limitations**.

- **Bug when removing a team member**:  
  After removing a member from the team, it's not possible to invite the same user again—or even a different one. This also appears to be related to a **membership plan limitation**.

- **Inconsistent behavior when modifying your own role**:  
  While you cannot remove yourself from the team (which is expected), it is possible to **downgrade your own role**, which can result in inconsistent permission behavior.

- **Pending members don’t have an ID**:  
  Since pending invitations lack a user ID, API calls for role changes or deletions can't be applied. While changing the role of a pending member may not make much sense, there **should be a way to cancel an invitation**, which is currently not supported by the API.

src/components/Organization/Team.tsx
```diff
@@ -94,6 +145,356 @@ export const useAllTeamMembers = () => {
...
+      </Popover>
+      {hasId && (
+        <>
+          <ChangeRoleModal isOpen={isRoleModalOpen} onClose={closeRoleModal} member={member} />
+          <RemoveMemberModal isOpen={isRemoveModalOpen} onClose={closeRemoveModal} member={member} />
+        </>
+      )}
+    </>
+  )
+}
```
elboletaire, line 496: These two components should be using the Menu component instead of popover. Also, use the isLoading from the Menu Button rather than returning a progress.

AlexanderArce:
> After trying to invite a member to my org, the team section crashes with the attached error imatge

Fixed

PR #1069, opened 2025-05-30
feat(memberbase): Create the new Memberbase UI

AlexanderArce:
…b to dashboard

Introduced new route `/admin/members` in the dashboard, added menu entry, routing handler and base layout with title and subtitle placeholders.


src/queries/organization.ts
```diff
@@ -226,7 +227,7 @@ export const useOrganizationSetup = () => {
     {
       id: SetupStepIds.memberbaseUpload,
       label: t('organization_setup.setup_steps.memberbase_upload', { defaultValue: 'Upload your memberbase' }),
-      to: '',
+      to: generatePath(Routes.dashboard.memberbase.members, { page: '1' }),
```
elboletaire, line 230: I don't think that `1` should be treated as a string, but as a number instead.

src/router/routes/dashboard.tsx
```diff
@@ -106,6 +109,37 @@ export const useDashboardRoutes = () => {
...
+                  <SuspenseLoader>
+                    <Memberbase />
+                  </SuspenseLoader>
+                ),
+                errorElement: <Error />,
+                children: [
+                  {
+                    index: true,
+                    element: <Navigate to={Routes.dashboard.memberbase.members} replace />,
+                  },
```
elboletaire, line 124: Why this navigate?

src/queries/members.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  return useMutation<AddMemberResponse, Error, Record<string, any>>({
+    mutationKey: QueryKeys.organization.members(organization?.address),
+    mutationFn: async (members) =>
+      await bearedFetch<AddMemberResponse>(fetchUrl, { body: { members }, method: 'POST' }),
+    onSuccess: () => {
+      setStepDone(SetupStepIds.memberbaseUpload)
+    },
+  })
+}
```
elboletaire, line 85: Given the API endpoint expects a list, and you're actually giving the possibility to pass a list of members, the hook (and related methods) should be plural (i.e. `useAddMembers`).

src/components/Memberbase/Members/Manager.tsx
```diff
@@ -0,0 +1,147 @@
...
+  member?: Partial<Member> | null
+}
+
+export const MemberManager = ({ control, member = null }: MemberManagerProps) => {
+  const { t } = useTranslation()
+  const toast = useToast()
+  const { isOpen, onOpen, onClose } = useDisclosure()
+  const btnRef = useRef(null)
+  const { columns } = useTable()
+  const addMember = useAddMember()
```
elboletaire, line 37: Pluralize to `addMembers` for coherence with its API.

src/components/Memberbase/Members/Manager.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  const errorToastMessage = isEdit
+    ? t('memberbase.edit_member.error', { defaultValue: 'Error updating member.' })
+    : t('memberbase.add_member.error', { defaultValue: 'Error adding member.' })
+
+  useEffect(() => {
+    if (member) {
+      methods.reset(member)
+    }
+  }, [member, methods])
```
elboletaire, line 67: Always add comments to effects explaining what are they doing.

elboletaire (changes requested):
I have not finished reviewing all of it, but I'll do a pause requesting changes so you can already work on them.

Other than the code comments, a few behavior/flow comments:
- Accessing the memberbase url redirects to a `/:page` url. The redirect itself is weird (I added a comment to that navigate asking why), but in case of redirecting, it should be properly setting the first page there
- Managing memberbase columns should store the configuration to local storage
- Both email and phone are not visible via the API (@emmdim emails should be visible, but phone might not be required given it's encoded). The phone would be encoded, so I don't think we should show it (nor as a configuration column nor as an actual column)
- Please set all non-working actions' buttons to `disabled` (like the export or the create group buttons)

elboletaire (changes requested):
Things that need to be addressed:
- [x] Memberbase pagination is not properly working. The url and buttons change state, but the query is not performed unless you load for the first time
- [x] The download of the template should include demo data, check [this](https://github.com/vocdoni/ui-scaffold/blob/develop/src/components/ProcessCreate/Census/Spreadsheet/index.tsx#L69-L87) for a currently working template (used in process create).
- [x] The email should be validated when adding members using the form
- [x] The phone field should be masked when the API returns any data there.
- [ ] There's no actual validation during imports (depends on https://github.com/vocdoni/saas-backend/issues/152)
- [x] There are new columns that should be added both to the form and the import process (at least BirthDate, Surname, NationalID)
- [x] Selecting a column during the import should disable it to avoid selecting it twice ([fig2](https://github.com/user-attachments/assets/4c5b1dfd-e597-441b-8126-6b4fd4959e7c))
- [x] There should be some space/margin here ([fig3](https://github.com/user-attachments/assets/3e3ff29a-22c9-4115-95b8-975d89fb32c9))
- [x] The usage given to the newly created MemberbaseContext could be perfectly handled by the outlet context, since you're using a layout for this specific behavior.

Also... I feel like this behavior of selecting `--NO IMPORT--` and having it as `Select column` is not very user friendly ([fig1](https://github.com/user-attachments/assets/fc829e14-3502-4ce3-8c5c-415b358fa4aa))

fig1: 
imatge

fig2:
imatge

fig3:
imatge

elboletaire (approved):

PR #1101, opened 2025-07-19
Replaced the free input with a stepper UI for setting the number of allowed vote overwrites.

ferranrego:
The value is constrained between 1 and 10. Added contextual helper text below the control to explain how it works.

### DIFF
--- a/src/components/Process/Create/Sidebar/ExtraConfig.tsx
+++ b/src/components/Process/Create/Sidebar/ExtraConfig.tsx
@@ -3,18 +3,74 @@ import {
   FormControl,
   FormErrorMessage,
   FormLabel,
-  NumberInput,
-  NumberInputField,
+  HStack,
+  Icon,
+  IconButton,
   Spinner,
+  Text,
   VStack,
 } from '@chakra-ui/react'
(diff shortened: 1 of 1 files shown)


src/components/Process/Create/Sidebar/ExtraConfig.tsx
```diff
@@ -3,19 +3,75 @@ import {
...
+        minW='48px'
+      />
+      <Text
+        flex='2'
+        textAlign='center'
+        py={3}
+        px={4}
+        borderLeft='1px'
+        borderRight='1px'
+        borderColor='gray.200'
```
elboletaire, line 53: Colors (in general) cannot be hardcoded, otherwise it's going to look bad on some of the themes

elboletaire (changes requested):
The selector looks good visually, but bigger than expected compared to other inputs

imatge

You're also hardcoding colors, which makes it look horrible on the dark theme:

imatge

Finally, chakra has some components to help you build this kind of interfaces:
https://v2.chakra-ui.com/docs/components/number-input

It would have been much better to use the number input provided by chakra and update its theme to fit our needs.

PR #1103, opened 2025-07-21
feat: Approval and Multichoice Election

AlexanderArce:


elboletaire (changes requested):
There are several logs in the code, either get rid of them or change them to be `console.info` is they're really needed.

We should also handle the abstention case here, if a user selects "minimum choices" to 0, we should be passing the required arguments to the SDK in order to create them with an abstention choice. You can see how it's handled during creation here: https://github.com/vocdoni/projects/blob/f4a19ea2b79e6cbc1c48757cba6d9dab0f3ab62f/packages/utils/src/process.ts#L127-L143

Note multi-choice and approval work slightly different here. Multi-choice have the `canAbstain` property, whilst approval votings do not, but we need to double-check if approval allows sending empty votes. In this case, it would mean that we must define the `canAbstain` and related attributes only for multi-choice elections.

elboletaire (approved):

elboletaire:
This has been accepted because it did comply with the acceptance criteria, though it surfaced some issues that we need to take a look at (I'll create a follow-up for this right away):
- Cannot abstain in approval votings
- When using multichoice and selecting a minimum and maximum, users cannot select the minimum unless there's an Abstain option and they select that too.

PR #1104, opened 2025-07-21
fix: improve blocker and fetchAccount

AlexanderArce:


src/components/Process/Create/index.tsx
```diff
@@ -441,16 +452,21 @@ export const ProcessCreate = () => {
...
   const description = methods.watch('description')
 
-  // Trigger confirmation modal when form is dirty and user tries to navigate away
   useEffect(() => {
-    const isBlocked = blocker.state === 'blocked'
+    const fetch = async () => {
+      await fetchAccount()
+    }
+    fetch()
+  }, [])
```
elboletaire, line 465: Add a comment explaining why is this effect here (I'd add it with an `IMPORTANT` to start with, since it's a fact).

Also, in these cases, simply use self-invoking anonymous function: 

```ts
  // IMPORTANT: We need to ensure the electionIndex is updated before creating an election.
  useEffect(() => {
    ;(async () => {
      await fetchAccount()
    })()
  }, [])
```

After adding the comment myself though, I realize that we could have issues if that person opens a new tab and creates a process in that new tab... we actually might need to do this fetch always BEFORE submitting (previously we were doing this after submitting, but I now realize that's a mistake).

elboletaire (changes requested):

PR #1128, opened 2025-08-22
fix: Prevent stuck blocker after successful form save

AlexanderArce:
This PR fixes an issue with the react-router blocker in the process creation form.

The blocker is intended to show a confirmation modal only when the form is dirty and the user attempts to navigate away. However, in some cases it also blocked navigation after a successful form submission or when navigating outside of the process creation flow, where no confirmation modal exists.

This caused navigation to get stuck with no way to proceed.

The blocker logic has now been refactored to ensure it only activates when the form is dirty and within the process creation context. Navigation after submit or outside the form now works as expected.


src/components/Process/Create/index.tsx
```diff
@@ -203,6 +220,76 @@ const TemplateConfigs: Record<TemplateTypes, TemplateConfig> = {
...
+  const proceed = () => {
+    isProceedingRef.current = true
+    closeAll()
+    blocker.proceed()
+    // limpieza defensiva en el siguiente tick
+    setTimeout(() => {
+      isProceedingRef.current = false
+      blocker.reset()
+    }, 0)
+  }
```
elboletaire, line 283: I think you should:
- ensure this method does not save the form, since right now the onfocus triggered by the form is storing in local storage and here we're not resetting it
- change the name of the method to something like `cancel` or even reuse the reset method when cancelling

elboletaire (changes requested):
Looks good, though right now clicking on "leave without saving" does not clear the form

PR #1129, opened 2025-08-22
feat: Update role permissions

### LINKED ISSUE #1062: Review the roles assignement and management under the Team Members tab
- [x] Admins can't change their role (For now, we'll improve this flow in the future).
- [x] The role "any" should disappear from the backend.
- [x] https://github.com/vocdoni/saas-backend/issues/130
- [ ] In the UI, these are the informative texts that should appear along with the roles. 
Image

- [ ] For the manager role: Under settings, it should see a greyed version of the tab "Organization details" and full access to the "Support" tab. The rest should not be visible.

The roles and its permissions are defined in [this document](https://www.notion.so/vocdoni/Vocdoni-App-management-1ed8cc4c1e428037bcb2fd78bc48abd2?pvs=4).

AlexanderArce:
closes #1062 
depends on https://github.com/vocdoni/saas-backend/pull/208


src/components/Organization/Team.tsx
```diff
@@ -265,8 +265,7 @@ const RoleRadio = ({ fieldName: title, description, value, isDisabled, ...props
       as='label'
       border='1px solid'
       borderRadius='md'
-      borderColor={isSelected ? 'gray.500' : 'gray.200'}
-      bg={isSelected && 'gray.50'}
+      borderColor={isSelected && 'gray.400'}
```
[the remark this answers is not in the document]
elboletaire: I understand this is the expected behavior, right @AlexanderArce? We only want the borderColor to be applied when it's selected. It's not entirely correct to use it this way in a style prop, since it will probably end up as a css property of `border-color: true`, but it's not going to cause any side effects trully, so it can be ignored.

src/components/Organization/Team.tsx
```diff
@@ -288,6 +287,19 @@ const RoleRadio = ({ fieldName: title, description, value, isDisabled, ...props
...
+  const hasOrgPermission = role.organizationWritePermission
+  const hasProcessPermission = role.processWritePermission
+
+  if (hasOrgPermission && hasProcessPermission) {
+    return <Trans i18nKey='role.full_access' defaults='Full access' />
+  } else if (hasOrgPermission || hasProcessPermission) {
+    return <Trans i18nKey='role.manage_members_votes' defaults='Manage members & votes' />
+  }
+
+  return <Trans i18nKey='role.read_only' defaults='Read-only access' />
```
elboletaire, line 300: Is it really `defaults` the prop for `defaultValue`? Because I'd say it's either `defaultValue` or using its children.

elboletaire (changes requested):

src/components/Organization/Team.tsx
AlexanderArce: `defaultValue` it's only when using `t()` when using `<Trans>` it's valid either `defaults` or the `children` https://react.i18next.com/latest/trans-component#alternative-usage-which-lists-the-components-v11.6.0

PR #1132, opened 2025-08-26
chore(deps): upgrade rainbowkit to v2

elboletaire:
Depends on https://github.com/vocdoni/ui-components/pull/219

### DIFF
--- a/.github/workflows/netlify.yml
+++ b/.github/workflows/netlify.yml
@@ -17,14 +17,6 @@ jobs:
         with:
           node-version: 22.x
 
-      - name: Cache Node.js modules
-        uses: actions/cache@v4
-        with:
-          path: node_modules
-          key: ${{ runner.os }}-node-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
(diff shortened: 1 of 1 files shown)


elboletaire:
Ok so, an update on this... The build works perfectly locally, but for some reason, the CI has an issue with an export from @noble/hashes:

image

Things I've tried (everything without success):
- Running it in different environments, including `act`
- Fixing versions for packages related to these exports, like `ox`
- Aliasing via vite that failing `/sha2.js` export to `/sha2` but then it fails with the error saying it cannot find `/sha2`
- Changing `moduleResolution` to other values, but it has to stay in `bundler` in order to avoid issues given by exports in the `ox` package using a very new format (using `node` fails with type errors from the `ox` package)
- Reviewed the versions in the CI and locally for packages like `node`, `yarn` and `vite` match.

Everything failed and I'm out of ideas right now... If I at least could try it locally, but I've not been able to reproduce this at all.

elboletaire:
OMFG it was due to the damn cache in github actions

PR #1183, opened 2025-09-04
docs: fix path to photo

GarmashAlex:
Now photo is shown in the file


elboletaire:
That path was OK. Note these markdown files are not created to be read/loaded via github. They are loaded in our app as static pages, see: https://app-dev.vocdoni.io/use-cases/ca/sports-clubs

PR #1219, opened 2025-09-15
chore: Voting page minor changes

### LINKED ISSUE #1216: Public voting page text and other minor changes
- [x] Voter's auth modal: replace "authentication required: participant not found in the census" by "The voters is not in the census or the credentials are not correct" (catch error and translate)
- [x] When the user already voted, autoscroll to "Your vote was successfully cast!"
- [ ] On the public-facing vote page, when the user already cast the vote, replace the text "Re-submit vote" to "Change my vote".

_friendly reminder to make each task on a single commit so they can be merged without squash and still maintain the history_

AlexanderArce:
closes #1216

### DIFF
--- a/src/components/Process/CSP/basics.tsx
+++ b/src/components/Process/CSP/basics.tsx
@@ -1,5 +1,6 @@
 import { useMutation } from '@tanstack/react-query'
 import { PublishedElection } from '@vocdoni/sdk'
+import { useTranslation } from 'react-i18next'
 
 export type CensusData = {
   authFields: string[]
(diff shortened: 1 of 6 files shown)


src/components/Process/CSP/basics.tsx
```diff
@@ -44,8 +46,17 @@ export const useTwoFactorAuth = <T extends number>(process: PublishedElection, s
         body: JSON.stringify(payload),
       })
       if (!response.ok) {
-        throw new Error((await response.json()).error)
+        const errorMessage = (await response.json()).error
+
+        const translatedMessage = /participant not found in the census/i.test(errorMessage)
+          ? t('csp.errors.participantNotFound', {
```
elboletaire, line 52: does not respect the naming, should be

`csp.errors.participant_not_found`, though I'd say to remove the `errors` if there aren't anymore.

src/components/Process/CSP/basics.tsx
```diff
@@ -44,8 +46,17 @@ export const useTwoFactorAuth = <T extends number>(process: PublishedElection, s
         body: JSON.stringify(payload),
       })
       if (!response.ok) {
-        throw new Error((await response.json()).error)
+        const errorMessage = (await response.json()).error
+
+        const translatedMessage = /participant not found in the census/i.test(errorMessage)
```
elboletaire, line 51: I think it'll be better if we check for the error number, not the error text. Don't we have a specific error number for this case?

elboletaire (changes requested):

src/i18n/locales/en.json
```diff
@@ -138,7 +138,7 @@
     "vote": {
       "abstain": "Abstain",
       "button": "Vote",
-      "button_update": "Re-submit vote",
+      "button_update": "Change my vote",
```
elboletaire, line 141: Should be changed in other languages too

elboletaire (changes requested):

PR #1387, opened 2025-11-17
chore: conditionally render group actions based on member count

### LINKED ISSUE #1386: [minor] hide create group (all) and remove (all) when we have zero items
From #1385 I do realize we can also hide the two "all" buttons in our memberbase when we have zero items. Let's do it, since it would improve the UX.

Image

AlexanderArce:
closes #1386

### DIFF
--- a/src/components/Memberbase/Members/index.tsx
+++ b/src/components/Memberbase/Members/index.tsx
@@ -308,14 +308,18 @@ const MemberFilters = ({ onDelete }: MemberFiltersProps) => {
           <IconButton size='xs' aria-label='search' type='submit' icon={<Icon as={LuSearch} />} />
         </InputRightElement>
       </InputGroup>
-      <CreateGroupButton includeAllMembers members={data?.members ?? []}>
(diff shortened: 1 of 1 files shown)


src/components/Memberbase/Members/index.tsx
```diff
@@ -308,14 +308,18 @@ const MemberFilters = ({ onDelete }: MemberFiltersProps) => {
...
       </InputGroup>
-      <CreateGroupButton includeAllMembers members={data?.members ?? []}>
-        {t('members.table.create_group_all', { defaultValue: 'Create group (All)' })}
-      </CreateGroupButton>
-      <Button leftIcon={<Icon as={LuTrash2} />} variant='outline' colorScheme='red' onClick={onDelete}>
-        {t('members.table.delete_all', {
-          defaultValue: 'Delete (All)',
-        })}
-      </Button>
+      {data?.members?.length >= 1 && (
```
elboletaire: It will work fine, but it would  have been better to use the pagination total items value, rather than the current page members length

PR #1480, opened 2025-12-23
F/subscription usage card

ferranrego:
This PR is a mockup that improve the Dashboard and includes a card to show usage and limits for Voting processes created, members existing in the memberbase and 2FA codes sent by email and by sms on the current plan of the user.

Note that this is still depending on the implementation of an endpoint to show the usage and enforce limits for the 2FA codes sent by email and sms https://github.com/vocdoni/saas-backend/issues/222


elboletaire:
superseded by #1507

Issue #1544, opened 2026-02-20
The CSP Auth login form has a weird line break due to a dot (ref  3)


elboletaire:
<img width="520" height="360" alt="Image" src="https://github.com/user-attachments/assets/c33f2d93-fd0b-44dc-af15-6d621471f329" />

elboletaire:
closed by #1547

AlexanderArce:
Apparently the problem was not resolved, reopening

elboletaire:
closed in #1552

Issue #1540, opened 2026-02-20
The confirm vote modal appears buggy before showing properly (ref 2)


AlexanderArce:

elboletaire:
closed by #1547

PR #1547, opened 2026-02-24
Fixed voting page modals

### LINKED ISSUE #1539: The  vote confirm modal is broken (ref 1)

AlexanderArce:
closes #1539 #1540 #1544


elboletaire (changes requested):
Avoid creating folders for single components. Also, for an svg component, it would be better having it on a different location. Move `components/Process/BallotBoxAnimated/index.tsx` to just `components/Layout/BallotBoxAnimated.tsx` (and do the same with its corresponding test).

Other than that LGTM

Issue #1537, opened 2026-02-19
Chakra v3 migration follow-up


elboletaire:
Feel free to edit this issue and add to the list whatever you find that looks wrong since the upgrade (you can use either stage or production for comparison btw):

- [x] On `/admin/processes/create` and on the voting process dashboards, the right sidebar doesn't extend to the bottom of the page. (ref 5)
- [x] On `/admin/settings/organization`, the organization type and country fields are not aligned. While we’re updating this view, we should also remove the duplicated "Other details" section and subtitle.
- [x] On `/admin/settings/support`, the title and subtitle have different style than on the rest of sections.
- [x] Replace the image of ref 1 with this asset: https://github.com/user-attachments/assets/9c4431c5-3fea-4a19-a3cf-224f06e59abc
- [x] Add copies (Pause, Cancel,...) to the buttons on the voting process dashboard (the buttons should always be visible but disabled when the process is finished) (ref 13)
- [x] Fix this title and subtitle so that it follows the same style as the rest of sections. (ref 12)
- [x] The button text in the informative left sidebar card is too long in some languages and breaks the button. Replace it with: Contactanos (ES), Contacta'ns (CAT), Contattaci (IT), and “Contact us” (EN). (ref 18)
- [x] When mapping a CSV to import members, the fields are not styled properly (ref 20)

| ref                                                 | image |
| ------------------------------------------- | ----- |
| 1 | ![ref 1](https://github.com/user-attachments/assets/fc4a4a63-ec9e-4b97-b57f-b5baa85b2efc) |
| 2 | ![ref 2](https://github.com/user-attachments/assets/c5dff8c6-69f9-47ee-ace8-475c81e2d341) |
| 3 | ![ref 3](https://github.com/user-attachments/assets/2e446006-6eea-457b-82b4-415214f91764) |
| 4 | ![ref 4](https://github.com/user-attachments/assets/262657ec-f76d-4145-a1e6-954ef1fadb7f) |
| 5| ![ref 5](https://github.com/user-attachments/assets/fc01b507-a6cc-4db0-a31d-581c12701aca) |
| 6| ![ref 6](https://github.com/user-attachments/assets/8fa6023d-9faa-49b3-8085-68739a1a6595) |
| 7| ![ref 7](https://github.com/user-attachments/assets/eef0c9bd-01b7-4993-907e-0f334c88e834) |
[...]

PR #1552, opened 2026-02-26
chore(i18n): correct punctuation to fix line break

AlexanderArce:
refs #1537


elboletaire (approved):
let's see how much does it take for someone to notice it and report it as an issue 😬

PR #1599, opened 2026-03-13
Improvementes for Chakra v3

### LINKED ISSUE #1590: Improvements and fixes Chakra v3
- [x] Cloning a voting process does not retain the voting method configuration (For example, if the original process is set to multiple-choice, the duplicated draft defaults to single-choice and selection limits are not copied)
- [x] The “rows per page” dropdown is hidden beneath the memberbase table container. #ref1
- [x] Ensure the spacing between the question description and the following element matches the spacing used between other elements in the same card. #ref3
- [x] The floating menu used to edit card descriptions (in extended info mode) appears underneath the card boundaries instead of above them. #ref4
- [x] Make sure that the cards match the style of the cards in the public-facing voting page (e.g. font sizes, the title must be in bold,...). This should apply to the rest of the public-facing voting page and the dashboard. Let's talk properly about it because we should work a bit more in deep so we have a final consistent version. I'll give more instructions. #ref5
- [x] Update the label on the “Extended information” cards by replacing “Project description” with “Description (optional)” across all supported languages.
- [x] The spreadsheet "upload again" button (in dashboard) should be outlined, not solid.
- [x] SpreadsheetAccess login button does not share styles with other login buttons (it's not filling the full width)
- [x] Confirm vote modal is weird in multichoice elections (looks good in all other cases).
- [x] On mobile devices, pagination looks broken with everything in a column, even though it should be in a single row.
---
**#ref1** 
Image
**#ref2**
Image
**#ref3**
Image
**#ref4**
Image
**#ref5**
Image

AlexanderArce:
closes #1590 

- Cloning a voting process does not retain the voting method configuration (For example, if the original process is set to multiple-choice, the duplicated draft defaults to single-choice and selection limits are not copied)

- Ensure the spacing between the question description and the following element matches the spacing used between other elements in the same card. #ref3

- Make sure that the cards match the style of the cards in the public-facing voting page (e.g. font sizes, the title must be in bold,...). This should apply to the rest of the public-facing voting page and the dashboard. Let's talk properly about it because we should work a bit more in deep so we have a final consistent version. I'll give more instructions. #ref5

-  The spreadsheet "upload again" button (in dashboard) should be outlined, not solid.

- SpreadsheetAccess login button does not share styles with other login buttons (it's not filling the full width)

- Confirm vote modal is weird in multichoice elections (looks good in all other cases).

-  On mobile devices, pagination looks broken with everything in a column, even though it should be in a single row.

### DIFF
--- a/src/components/Pagination/PaginatedTableFooter.test.tsx
+++ b/src/components/Pagination/PaginatedTableFooter.test.tsx
@@ -1,0 +1,88 @@
+import { render, screen } from '~src/test-utils'
+import type { ReactNode } from 'react'
+import RoutedPaginatedTableFooter from './PaginatedTableFooter'
+
+const mockUseRoutedPagination = vi.fn()
+
+const serializeResponsiveProp = (value: unknown) => {
+  if (value == null) return ''
+  return typeof value === 'string' ? value : JSON.stringify(value)
+}
+
+vi.mock('@chakra-ui/react', async () => {
+  const actual = await vi.importActual<typeof import('@chakra-ui/react')>('@chakra-ui/react')
+  const React = await vi.importActual<typeof import('react')>('react')
+
+  return {
+    ...actual,
+    Box: ({
+      children,
+      display,
+      flexDirection,
+      flexWrap,
+      alignItems,
+      justifyContent,
+      ...props
+    }: {
+      children?: ReactNode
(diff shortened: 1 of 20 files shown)


emmdim (approved):
LGTM

PR #1615, opened 2026-04-17
feat(dashboard): Add Voting Report PDF functionality

### LINKED ISSUE #1525: Implement downloadable vote reports for compliance (PDF Report)
Specs soon

AlexanderArce:
- Implemented VotingReportPdfButton and VotingReportPdfMenuItem components for downloading election reports in PDF format.
- Created VotingReportPdf integration and unit tests to ensure proper functionality and error handling during PDF generation.
- Updated ProcessesTable and ProcessView components to include PDF download options.
- Enhanced existing tests to verify the presence of PDF download actions in relevant components.
- Added localization strings for PDF report features in multiple languages (English, Spanish, Catalan, Italian).

closes #1525


elboletaire (changes requested):
A part from reviewing the copilot issues, which seem legit (specially the UTC one), there are two things that should be improved:
- The "download PDF" button styles in the process view differs from its sibling button (show in explorer).
- We should have some page breaks (they're called page wrappers in react-pdf), so contents don't break the way they do. One should be defined up to the point 3, this way the point 4 would appear in the next section. Same for the legal notice, at the end of the PDF.

elboletaire (approved):

PR #1632, opened 2026-04-30
Prod 2.5.2 upgrade

elboletaire:
Notable changes since 2.5.1:

- improved 2fa voting modal
- added resend flow for 2fa codes
- `canAbstain` specific logic removed for now
- Crisp chat only loads after accepting cookies (only for crisp chat enabled deployments ofc)
- UX improvements due to the recent auto-groups addition
- PDF reports for votings having results


elboletaire:
Note, this is NOT 2.5.2... after this crazy week with cherry-picking and breaking the git graph on multiple levels... this should probably be 2.6.0 already, but we'll rearrange git branches next week.

PR #1643, opened 2026-05-12
chore: UI Voting Page fixes

### LINKED ISSUE #1630: UI issues on public voting page (modal alignment, incorrect message, selection behavior)
- Remove the dashboard button from the public voting page (even when you're an admin)

- The modal should be centered on the screen (vertically and horizontally) with sufficient horizontal spacing to prevent it from appearing attached to the edges. (image 1)
- In the copies under the OTP fields, a period is missing at the end of the sentence. (image 1)

Image

AlexanderArce:
closes #1630

### DIFF
--- a/src/components/Navbar/index.test.tsx
+++ b/src/components/Navbar/index.test.tsx
@@ -35,4 +35,26 @@ describe('Navbar', () => {
 
     expect(screen.getByRole('link', { name: 'Login' })).toBeInTheDocument()
   })
+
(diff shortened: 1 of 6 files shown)


elboletaire (commented):
Not sure a toggle was required here since the voting page is the only publicly available page (other than the organizations page, which is kinda hidden), but for the sake of reducing the PR burden, I'll accept it.

PR #1647, opened 2026-05-15
fix(spreadsheet): trim whitespace from all string cells on memberbase parse

### LINKED ISSUE #1644: Uploading a memberbase with spaces trimming issue
Uploading a memberbase with spaces in fields like email and membernumber the spaces are not trimmed

argos-code:
## Summary

When uploading a memberbase file (CSV or XLSX), leading and trailing whitespace in any string cell was preserved verbatim. This caused imports to fail or store corrupted values when the source file contained accidental spaces in fields like `email` or `membernumber`.

The fix applies `.trim()` to every string-typed cell immediately after the xlsx worksheet is parsed, before any header extraction, validation, or storage.

## Linked issue

Closes #1644

## Changes

- **`src/components/Spreadsheet/SpreadsheetManager.ts`** — `getSheetsData()` now maps over every cell and trims string values. One-line change at the point where raw sheet data becomes `string[][]`.
- **`src/components/Spreadsheet/SpreadsheetManager.test.ts`** *(new)* — Unit test that constructs an xlsx workbook with padded headers and data cells, injects it directly into `SpreadsheetManager` (bypassing `FileReader`), and asserts the parsed header and data are trimmed.

## Tests run

```
pnpm test

Test Files  113 passed (113)
      Tests  425 passed | 2 skipped (427)

pnpm lint

tsc && prettier --check src
All matched files use Prettier code style!
```

## Risks and review focus

- The trim is applied generically to **all** string cells, not only `email` and `membernumber`. This is intentional — any future column benefits automatically — but reviewers should confirm this is acceptable for fields where internal whitespace is theoretically meaningful (none are known).
- Non-string cells (numbers, dates produced by xlsx with `raw: false`) pass through unchanged.
- No behaviour change for well-formed files; the existing test suite (113 files) continues to pass.

## Notes for maintainers

The test bypasses `FileReader` (which is not available in jsdom) by calling the private `getSheetsData` method directly via `(manager as any)`. This matches the only practical way to unit-test this class without a real browser File API.


p4u:
@argos-code amend in a single commit and remove unrelated files from the PR

PR #1654, opened 2026-05-21
feat(process): show authenticated voter label in navbar

### LINKED ISSUE #1650: When a voter authenticates, show in the UI an identifier of that member
We should make it clearer to voters that they are authenticated.

### Requirements

- For **CSP** and **Spreadsheet** authentication methods:
  - Display the value of the first credential field used during authentication.
  - We assume that the displayed value selected by the organization will typically be something identifiable to the voter, such as:
    - Name
    - Member ID
    - National ID
    - etc.

- For **Web3** authentication:
  - Do not display any additional identifier.

### UI Notes

- Use the default styling system.
- The identifier should appear inside a visually highlighted container/box so it is easy to notice.
- The attached screenshot is only a visual reference and should not be replicated literally.

Example reference:

Image

AlexanderArce:
closes #1650

### DIFF
--- a/src/components/Navbar/index.test.tsx
+++ b/src/components/Navbar/index.test.tsx
@@ -35,4 +35,53 @@ describe('Navbar', () => {
 
     expect(screen.getByRole('link', { name: 'Login' })).toBeInTheDocument()
   })
+
+  it('renders the authenticated voter label instead of the dashboard button', async () => {
+    const i18nInstance = await createTestI18n({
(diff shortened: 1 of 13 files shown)


elboletaire:
@AlexanderArce please to rebase the branch with develop

AlexanderArce:
moved to https://github.com/vocdoni/vocdoni-app/pull/1670

PR #1681, opened 2026-06-02
Added a new section in the voting process dashboard for organizations to check who voted, from the list of participants

ferranrego:
## Acceptance Criteria

- [x] When a member is edited in the Member Base, the participant verification list is updated accordingly.
- [x] Member search works using the same authentication fields configured for the voting process, the organizer can select which one to use to search.
- [x] Member search clearly indicates whether a voter has already voted or not.
- [x] If a member is not found in the Member Base, an informative message is displayed to the user.


elboletaire (approved):
LGTM though next time you should ensure the AI adds tests for new functionalities.

PR #1682, opened 2026-06-02
Stage cherry-pick hotfix

elboletaire:
Currently working on it, we have to bring to stage (and prod) some of the dev features without publishing SSR yet:

- [x] All the pdf related changes (at least #1525, but needs merge first ofc)
- [x] Be able to search people on census #1681 
- [x] Auto-close CSP sessions #1680
- [x] Trim memberbase emails #1667
- [x] Remove dashboard button in process view layout


ferranrego:
Just so we don't forget, we also need to merge the feature that makes the authentication form fields case-insensitive, at least for the national ID number and email fields, but perhaps for name, surname and member number would be useful too.

elboletaire:
> Just so we don't forget, we also need to merge the feature that makes the authentication form fields case-insensitive, at least for the national ID number and email fields, but perhaps for name, surname and member number would be useful too.

This has NOT been fixed in develop. And it requires further development, not just what we did in `d/ustec` (we must ensure both the form and the import processes do actually convert to lowercase the emails, we said to do this part in the backend directly though cc @emmdim).

ferranrego:
This is one of the main issues with our client, as they had some eligible voters having problems with that (they use the email to login with 2FA, is it feasible to implement by tomorrow?

elboletaire:
> This is one of the main issues with our client, as they had some eligible voters having problems with that (they use the email to login with 2FA, is it feasible to implement by tomorrow?

Everything is possible my friend... the question is, will elboletaire's health be fine after this week's workload? we'll see 👀 

(AKA I'll do my best to also add it tomorrow)

ferranrego:
> > This is one of the main issues with our client, as they had some eligible voters having problems with that (they use the email to login with 2FA, is it feasible to implement by tomorrow?
> 
> Everything is possible my friend... the question is, will elboletaire's health be fine after this week's workload? we'll see 👀
> 
> (AKA I'll do my best to also add it tomorrow)

ahaha I’m excited to see which one survives first: the feature or elboletaire  👀

PR #1712, opened 2026-07-09
refactor(auth): remove VocdoniSDK and use the new API

elboletaire:
## Context

First step of migrating the app off \`@vocdoni/sdk\` (RemoteSigner) and \`@vocdoni/react-components@1\` onto the new **integrator-sdk** monorepo (\`@vocdoni/api-client\`, \`@vocdoni/api-types\`, \`@vocdoni/api-voting\`, \`@vocdoni/react-providers\`, \`@vocdoni/react-components@2\`).

**North star:** remove all client-side Vochain API interaction. Everything the admin app needs — auth/sessions, the active organization, org metadata — now goes through the Vocdoni SAAS backend instead of reading from the Vochain.

This PR is **Step 1: remove the RemoteSigner** from the admin (client-rendered) surface. It's intentionally scoped; later steps are listed below.

## What changed

- **integrator-sdk client provider** — mounts the new \`VocdoniApiClient\` (via react-providers \`ClientProvider\`) for the authenticated app, exposed as \`useApiClient()\` to avoid clashing with the legacy \`useClient()\`. Links the integrator-sdk packages via pnpm \`link:\` and dedupes React in Vite.
- **Auth / session** — the token lifecycle (token, expiry, persistence, refresh, logout) is delegated to the react-providers \`AuthProvider\`. The active organization address, previously resolved through the SDK \`RemoteSigner\`, now comes from the SAAS API (\`auth.addresses()\`), with the selection persisted for multi-org accounts. OAuth (Google) sessions are injected via \`setSession\`; guards and the org switcher re-resolve via \`refreshAddresses\`.
- **Org data** — the session's active organization is sourced from the SAAS API (\`organizations.get\`) through \`useSaasAccount\`. Every \`useClient().account?.address\` read across actions, memberbase, organization, process, pricing and dashboard surfaces is replaced with \`useAuth().currentAddress\`. Org create/edit go through the SAAS API (\`provisionAccount\`, \`PUT\`) instead of on-chain account transactions.
- **Admin election reads + lifecycle** — process listing (\`dashboard/processes/all\`, \`ended\`), organization dashboard (\`Votings\`), and election status changes (publish / pause / end / cancel) are fully sourced from SAAS via the integrator-sdk \`elections\` client. The \`toPublishedElection\` adapter bridges the SAAS \`Election\` type to the \`PublishedElection\` shape the view components still consume.
- **Pagination** — fixed for api-client's 1-based page convention (page 1 = first page); \`lastPage\` is now the total page count fed directly to \`RoutedPaginationProvider\`.
- **Providers cleanup** — removed the outer \`ClientProvider\` that was passing legacy \`env/signer/options\` props that the v2 provider doesn't accept. Replaced with a bare \`ToastProvider\`. The \`environment\` export (v1-only) was removed from \`Aside.tsx\` — it was causing a native ESM \`SyntaxError\` at browser module load time.
- **Tests** — a global \`useAuth\` mock drives \`currentAddress\` and the SAAS session in place of the former RemoteSigner; \`AllProviders\` wraps the new client + auth providers.

### Deliberate seam (not backwards-compat)

\`SaasAccountProvider\` adapts the SAAS \`OrganizationInfo\` onto the app's \`.account\`-nested shape. This is **not** old-SDK compat — the data is 100% SAAS. The org display components (\`Organization/Header\`, \`Process/View\`, \`CreatedBy\`, \`LegalNotice\`) are shared with the **public SSR pages**, which still feed the nested shape from the Vochain path. The adapter keeps both contexts working and is deleted once the public/SSR path also moves to SAAS (see below).

### Deferred (TS errors suppressed / files intentionally skipped)

- **Voting flow** — \`Aside\`, \`LogoutButton\`, \`SharedCensus\`, \`CSP/\` still consume \`useElection()\` from \`@vocdoni/react-components\` (unchanged — that component surface hasn't moved yet).
- **Election creation** — \`Create/index.tsx\` still uses \`electionService\`/\`createElectionSteps\` from the old SDK (blocked on a SAAS create endpoint).
- **Public / SSR pages** — Vike loaders still call the legacy Vochain path (blocked on an address→Mongo-ID reverse lookup in the SAAS API).

## Verification

- \`pnpm test\` → green
- \`pnpm lint\` (tsc) → clean on non-deferred files
- Prettier → clean on changed files
- Dev server boots cleanly; admin election list, dashboard, and lifecycle actions go through SAAS.

## Planned next (not in this PR)

1. **Public / SSR org path** — migrate \`src/ssr/public-pages.ts\` + Vike loaders to SAAS, flatten the shared components to react-providers \`useOrganization\`, and **delete the \`SaasAccountProvider\` adapter**. (Blocked on address→Mongo-ID reverse lookup in SAAS API.)
2. **Voting flow** — migrate \`Aside\`, \`LogoutButton\`, \`SharedCensus\`, \`CSP/\` to integrator-sdk.
3. **Election creation** — migrate \`Create/index.tsx\` once the SAAS create endpoint is available.
4. Full \`@vocdoni/sdk\` removal and dead-code cleanup (\`utils/pagination.ts\`, \`VocdoniClientProvider.tsx\`, \`walletClientToSigner\`).

> Note: depends on the linked integrator-sdk branch (\`feat/session-renewal-and-org-info-types\`: \`setSession\`/\`expiry\` in react-providers, expanded \`Organization\` type in api-types).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


elboletaire:
@argos review

elboletaire (commented):

Thanks — six of the twelve findings were real and are now fixed. The other five are pointing at code this PR never touches, and one had already been fixed locally before the review ran.

Before the details, one process note: **you reviewed a 384-file diff, but this PR is 219 files against its base** (`1782473d`). That over-wide diff is what produced all five false positives — every one of them is in a file where `git diff 1782473d..HEAD -- <file>` is empty. Worth tightening, because #1 in particular was reported as a Major regression on the strength of a `git show` result that doesn't reproduce.

---

## Fixed

**#2 · `Process/Create/index.tsx` — publish orphaning the draft** → 00f06e4f
Confirmed exactly as described, including the second defect (created id never written back, leaking a process per failed publish). Both writes now go through a single `writeDraft` helper that resolves create-vs-update inside the queue against `draftIdRef` and records the id before returning. Pinned by a regression test that fails with the decision moved back into the render closure (`create` called twice).

**#3 · `certificate-data.ts` — weighted multichoice inflation** → fbaa0ff0
Confirmed. The decoded histogram has one field per pick slot and every ballot fills all of them, so the column total is `maxCount x cast power`. Divided by the question's pick slots now, and added the weighted-multichoice case (your 7/3-of-20 scenario) that the suite was missing — it fails with the divisor removed.

**#4 · `useAuthProvider.ts` — transient failure reads as "no organizations"** → 8cce5478
Confirmed end-to-end through to `NoOrganizationsPage`. Now retries anything that isn't a 4xx (a 4xx is a definitive answer, not a failure to get one), exposes `addressesError`, and `OrganizationProtectedRoute` renders a distinct unavailable state with a retry action instead of onboarding.

**#8 · `draft-mapping.ts` — extended-info toggle lost** → 8b50258d
Confirmed, including the JSON-drops-undefined mechanism. Keyed off metadata presence. Two existing cases asserted the old content-based inference and were flipped, since stored metadata means the toggle was on.

**#10 · `queries/integrators.ts` — `as string` casts** → 63ecb8db
Confirmed. Real guards instead of the casts, and the create/revoke controls are disabled while no address is selected.

**#11 · `i18n/index.test.ts` — test didn't pin the deep merge** → 4c7c1993
Confirmed, and verified directly: with `deepMergeResources` reverted to the shallow spread, the old `statuses.ready` assertion still passed. Now asserts `question_types.budget_title` / `quadratic_title` and `validation.choices_range` — SDK-only keys whose parent object the app overrides for other siblings — which do fail under a shallow spread.

**#7 · `applySelection` raw string equality** → afab63c9
Already fixed locally before your review ran; now committed. Good catch regardless.

---

## Rejected — pre-existing, not introduced here

Each verified with `git diff 1782473d..HEAD -- <file>` (empty) and `git show 1782473d:<file>`.

**#1 · Fallback language flips to `ca` — the supporting evidence doesn't reproduce.**
The review states: *"I confirmed `git show develop:src/i18n/languages.ts` starts `en`"*. It starts `ca`, on `develop` and at the merge-base. `languages.ts` and `app-env-build.ts` are untouched by this PR, and `Providers.tsx:161` (`fallbackLanguage: supportedLanguages[0]`) is byte-identical at the base. Nothing here alphabetizes anything. The underlying observation — that deriving the fallback from map insertion order contradicts `DEFAULT_FALLBACK_LANGUAGE = 'en'` — is fair and worth a follow-up, but it is not a regression from this PR.

**#5 · `OrganizationTypeGuard.test.tsx` vacuous — correct, but not "the new guard".**
The analysis is right: the static `/integrators` route outranks the `*` splat, so the guard never mounts and `getByText` is satisfied by the sibling div. Both the guard and its test are unchanged by this PR. Worth fixing, separately.

**#6 · Dockerfile build-arg contract dead — the migration predates this branch.**
`Dockerfile` is unchanged, and `vite.config.ts` had no `define`/`__APP_ENV__` block at the base either. The build-time-to-runtime move already landed before this branch, so the dead ARG and stale README aren't this PR's drift.

**#9 · `Integrator/Overview.tsx` error branch — unchanged file.**
The point stands on its merits (`!data?.limits` shows a hard error for a valid `enabled: false`), but the file isn't touched here.

**#12 · Stale `importmeta.d.ts` — unchanged file.**
Correct that only `DEV`/`SSR`/`BASE_URL`/`VITEST` are read, and identical (45 lines) at the base.

I'll open a follow-up issue for #1, #5, #9 and #12, which are genuine even though they don't belong to this PR.

---

## Verification

`pnpm test` 604 passed / 1 skipped · `tsc --noEmit` clean · `prettier --check` clean · `pnpm translations` idempotent. New locale keys translated for `en`/`es`/`ca`/`it` per the repo's i18n rules. Each of the four behavioural fixes has a test verified to fail with the fix reverted.

Also worth noting the coverage gap you flagged — SSR/Vike pages, `ApiClientProvider` request-scoping, `pdf-document.tsx` and the theme changes — still stands; nothing in this round touched it.

PR #1713, opened 2026-07-09
fix(dashboard): correct title spacing, casing, overflow and link underlines

ferranrego:
## What

Fixes a set of UI issues on the dashboard home screen, centred on the **Recent voting processes** card but applied globally where the root cause was shared:

- **Title casing** — sentence-case the English titles (`Recent voting processes`, `Quick actions`); other locales already complied.
- **Card overflowing the viewport** — removed `flexWrap='wrap'` from the column-flex `DashboardBox` (wrapping made children size to their content and spill past the card) and added `minW={0}` so inner `truncate` actually engages. Quick-action buttons now truncate their labels on narrow screens instead of pushing the layout.
- **Inconsistent hover underline on titles** — the `link` recipe `plain` variant no longer fades/shifts the underline on hover (it was fighting the base styles via Chakra's default variant), and a new `unstyled` variant is used for process-title links so they never underline.
- **Tiny spacing between wrapped title lines** — the `text` recipe size variants now use **unitless** line-heights, so any `fontSize` override degrades to a proportional ratio instead of pairing with a stale pixel value. Card titles now use the proper `size` variant rather than an inline `fontSize` override. `ElectionTitle` line-height raised to `1.3` for theme consistency.

## Why

The card used `flexWrap='wrap'` on a column flex, a fixed-pixel line-height paired with an overridden font size, a permanently-underlined `link` base colliding with Chakra's default hover styles, and hardcoded Title Case in the English locale. These produced the reported overflow, cramped multi-line titles, flickering underlines, and wrong capitalization.

## How verified

- `pnpm lint` — clean (tsc + prettier).
- `pnpm test` — 514 passed, 3 skipped.
- Live preview on the authenticated dashboard at 315px, 375px and 1280px: zero horizontal overflow, healthy title line spacing, long process titles truncate, quick-action labels ellipsize when space runs out, and process titles never underline (on hover or otherwise) while action links keep a consistent solid underline.


src/components/Account/Edit.tsx
```diff
@@ -19,7 +19,7 @@ export const AccountEdit = () => {
         <AccountForm profile={profile} />
       </DashboardBox>
       <DashboardBox p={6}>
-        <Text fontSize='2xl' fontWeight='600'>
+        <Text size='2xl' fontWeight='600'>
```
elboletaire, line 22: "size" acts different than "fontSize". Is this change really intended? Note using "size" changes the font size depending on the device/screen

src/i18n/locales/en/common.json
```diff
@@ -234,11 +234,11 @@
       "how_first_vote": "How to create your first vote",
       "no_organization": "No organization found",
       "no_subscription": "No subscription available. Please create an organization to activate your plan.",
-      "quick_actions": "Quick Actions",
+      "quick_actions": "Quick actions",
       "recent_voting_description": "Your organization's latest voting activities",
-      "recent_voting_title": "Recent Voting Processes",
+      "recent_voting_title": "Recent voting processes",
```
elboletaire, line 239: German needs updating too

src/i18n/locales/en/common.json
```diff
@@ -234,11 +234,11 @@
...
       "subtitle": "Here's an overview of your organization's voting activities",
-      "title": "Here's an overview of your organization's voting activities"
+      "title": "Dashboard"
```
elboletaire, line 241: Review other languages too please, Greek, Italian and German also has some long string here.

elboletaire (changes requested):

elboletaire (approved):

PR #1716, opened 2026-07-13
feat(ui): shared UI kit — ConfirmDialog, EmptyState (B1)

ferranrego:
## Summary

First item of **Vertical B** (UI kit & component architecture). Starts the real `components/ui` kit so features compose primitives instead of rebuilding modal/empty-state markup by hand.

> **Stacked on [#1715](https://github.com/vocdoni/vocdoni-app/pull/1715)** (Vertical A). Base is `fix/theme-vertical-a` so this diff shows only B1. GitHub will retarget it to `develop` automatically once #1715 merges.

## Two primitives

| Primitive | Replaces | Notes |
|---|---|---|
| **`ConfirmDialog`** | The identical hand-rolled `Flex` + outline Cancel + red confirm footer that every `DeleteModal` call site repeats (8 sites) | Declarative destructive-confirm with a built-in footer. **Coexists** with the imperative `useConfirm`/`ConfirmActionModal` path (left untouched) — the declarative pattern is needed when the confirmed action is async and the dialog must stay open (with a spinner) until `onSuccess`. |
| **`EmptyState`** | Near-identical `NoElections` / `NoOrganizations` / `NoResultsFiltering` cards + Team's inline `UsersEmpty` | Icon-or-illustration + title + description + optional CTA. |

## Scope (deliberately bounded)

B1 = **create the primitives + migrate proof consumers**. A `Field` primitive was prototyped and dropped — it saved no mental model while leaking the underlying Chakra API through `*Props` escape hatches. The existing `FieldRoot as FormControl` alias pattern is the right level of abstraction for form fields.

## What's migrated as proof

- `NoElections`, `NoOrganizations`, `NoResultsFiltering`, Team `UsersEmpty` → `EmptyState`
- Team's two hand-rolled `DeleteModal` confirms → `ConfirmDialog`

## Design decisions (confirmed with maintainer)

- **Confirm modals coexist** (declarative `ConfirmDialog` vs imperative `useConfirm`) rather than unifying — the two call styles serve different needs. `useConfirm` closes the dialog immediately on click; `ConfirmDialog` keeps it open during the async mutation and closes in `onSuccess`.
- **File moves deferred** — new primitives land in `ui/`; existing files rebuilt in place.

## Verification

- `pnpm lint` (tsc + prettier) clean; full `pnpm test` green (564 passed).
- Colocated tests added for both primitives (4 tests).

## Standardizations to eyeball (need a logged-in dashboard)

Minor, intentional consistency changes on the migrated empty states: Team's zero-state icon 80px→48px (standard size), `NoResultsFiltering` heading `md`→`lg`, and empty-state body text now centered. Fallback is trivial if any reads wrong.


src/components/ui/ConfirmDialog.tsx
elboletaire, line 1: The AI here completely missed an already existing implementation. `@vocdoni/react-components` has a ConfirmModal component and hooks that we're already using in many places and can be 100% customized via the recipe and the component definition.

elboletaire (changes requested):

elboletaire (approved):

PR #1719, opened 2026-07-14
chore(stress): add HTTP load-test harness for the SSR server

elboletaire:
Adds stress/loadgen.mjs (dependency-free closed-loop HTTP load generator) and
stress/run-stress.sh (builds the image, runs it under CPU/memory limits, ramps
concurrency until the server degrades, and reports the breaking point). Run
artifacts under stress/results/ are gitignored.

The build context was ballooning to 16GB+ because .worktrees (29GB) was not
ignored, making image builds unusably slow. Also exclude .claude, .netlify
and stress/results.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,4 +7,8 @@ coverage
 .codex
 .agents
 .pi
+.claude
+.worktrees
+.netlify
+stress
 skills-lock.json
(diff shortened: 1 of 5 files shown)


elboletaire (commented):
✅

PR #1790, opened 2026-09-15
feat(home): add HOME_PROCESS_ID to serve a process as the app root

elboletaire:

## What

Adds a `HOME_PROCESS_ID` env var holding a single process id. When it is set, the app root renders that
process' voting page — the same view served at `/:lang/processes/:id`. When it is unset, the root behaves
exactly as today.

Both the bare root (`/`) and every localized root (`/en`, `/es`, …) render it, so the page is what a
visitor gets before *and* after the client-side language redirect.

### Why this name

Runtime env vars in this repo are unprefixed — they are resolved on the server from `process.env` in
`buildAppEnv()`, stored on Vike's `globalContext` and forwarded to the client via `passToClient`, so a
single Docker image stays configurable with `docker run -e ...`. There is no `VITE_` prefix anywhere
(that mechanism was deliberately removed), so `HOME_PROCESS_ID` sits next to `PROCESS_IDS`,
`PRIMARY_COLOR` and friends unchanged.

## How

Wired at the Vike page/routing/SSR level, not as a client-side redirect, so the root is real
server-rendered HTML:

- `src/pages/home-process/` — a new SSR page whose **route function** claims `/` and `/:lang` when the
  env var is set. Route functions ignore filesystem routing, which is what lets one page own both roots.
  `+Page`/`+Head`/`+title`/`+description`/`+lang` re-export the same shared modules the process page uses.
- `getHomeProcessRouteMatch()` in `src/ssr/public-routes.ts` — the matcher, kept free of runtime imports
  like the neighbouring matchers (Vike executes `+route.ts` with plain Node, outside Vite's resolver).
- `loadHomeProcessPublicPageData()` — the existing process loader with a different path builder, so
  `+data` is unchanged in substance.
- The localized SPA catch-all gives up the root by calling **that same matcher** rather than re-deriving
  the condition. Two route functions matching one URL is ambiguous, and a divergence between the two
  checks would be a routing failure that only appears when the env var is set.

### Metadata

Title, description, Open Graph, Twitter card and JSON-LD all describe the process. Canonical URL and
`hreflang` alternates point at the **roots** (`/en`, `/es`, …), not at `/:lang/processes/:id` — those
alternates also feed the navbar language switcher, so pointing them at the process page would have
dropped visitors off the homepage every time they switched language.

### One fix that was required, not incidental

`AnnouncementBanner` reads its dismissed flag through `useLocalStorage`, which **throws** when rendered on
the server. It only renders at `Routes.root`/`Routes.plans`, and neither was ever server-rendered before —
so this bug had been latent, and the new SSR root hit it immediately with a 500 on every request. The
storage-backed part is now mounted only after hydration. Nothing is lost by skipping the server pass: the
dismissed flag exists only on the client, so a server-rendered banner would flash for everyone who had
already closed it. `AnnouncementBanner.test.tsx` pins this with `renderToString` — verified that the test
fails with the fix reverted.

## How it was verified

`pnpm lint` and `pnpm test` are green (927 passed, 1 skipped). `pnpm translations` produces no diff — the
page reuses the process view's existing copy, so no new keys.

Manually, against a real archive-era process
(`6b342d99…000001`, `VOCDONI_ENVIRONMENT=prod`), on **both** the Vite dev server and a production build
served by `pnpm serve:ssr`:

**Env set**
- `/`, `/en`, `/ca` → `200`, server-rendered process page (confirmed in a real browser: the process view
  renders, hydrates without error, and the URL stays on `/` — no redirect loop).
- `<link rel="canonical" href="…/en">`, alternates on `…/ca`, `…/es`, …, `<html lang="en">`, and the
  process/organization name in the title, OG and JSON-LD.
- `/en/processes/:id`, `/en/processes/:id/summary`, `/ca/organization/0x…`, `/en/plans`, `/en/admin` all
  still `200` — nothing else changed hands.

**Env unset** (same build, same image, only the variable removed)
- `/` → the SPA shell with the default title and *no* canonical tag, then the usual client redirect to
  `/en` and on to the existing homepage behaviour (`→ /admin → signin` for a logged-out user with no
  `CUSTOM_ORGANIZATION_DOMAINS`/`PROCESS_IDS` match, which is what `src/router/routes/home.tsx` already
  did).
- `/en`, `/en/plans` unchanged.

**Bad input** — with a typo'd id, the root returns a proper `404` page, not a `500`. The id is only
checked for emptiness at config time; whether it resolves is answered by the API per request, exactly
like any other unknown process id.

## Known follow-up (deliberately out of scope)

The SSR cache in `server/ssr-cache.mjs` matches `/:lang/(organization|processes)/:x` only, so the root is
**not** cached while `/:lang/processes/:id` is (1 min TTL). A deployment that sets `HOME_PROCESS_ID` is by
definition putting its highest-traffic page at the root, so it would hit the API on every request. Worth a
follow-up that extends the cacheable pattern to the roots when the var is set; I left it out to keep this
diff to the feature.


elboletaire:
imatge

PR #1793, opened 2026-09-16
chore(process): trim the voting view sidebar

elboletaire:

## What

Two cuts to the info panel on the public voting view (`src/components/Process/View.tsx`):

1. **Anonymity card loses its second paragraph.** It rendered two sentences: the mode
   (`process.anonymity.{anonymous,private}.description`) and the mechanism behind it
   (`process.anonymity.mechanism.blind_csp.description` — "The credential service provider
   authorized each ballot with a blind signature…"). Only the first stays.
2. **Voting method card removed entirely.**

## Why

The blind-signature explanation is detail for whoever is auditing the process, not for a voter
glancing at the sidebar. It is unchanged and still shown in the organizer dashboard
(`Dashboard/ProcessView.tsx`) and in the PDF report (`VotingReportPdf/certificate-data.ts`), so
the translation key is kept.

The voting method card is plain wrong now that processes are multi-question: one process can
carry a single-choice, a multi-choice and a ranked question at once, and the card read the
ballot type of `questions[0]` and presented it as the method for the whole thing. Removing it
for now rather than trying to render a list.

## Fallout

- `useVotingMethodLabel` in `resultTypeLabels.ts` had no other consumer, so it goes too.
  `useResultTypeLabel` stays — the dashboard uses it.
- `pnpm translations` dropped the three keys that became unreferenced, across all 10 locales:
  `process.voting_type`, `process.voting_method.unknown`, `process.voting_method.weighted_format`.
- `View.test.tsx` had a case asserting the blind-signature paragraph *was* rendered; it now
  asserts the opposite. The single-match pin on the mode sentence — which
  `e2e/anonymous-voting.e2e.ts` relies on via a strict `getByText` — is kept and still holds.

## Worth a second opinion

`weighted_format` ("Weighted {{ base, lowercase }}") was the only place the public page signalled
that a process uses weighted voting. That nuance now appears nowhere on the public side; the
dashboard's `useResultTypeLabel` gives the bare ballot type without it. If weighted processes
should still say so somewhere, it needs a new home.

## Verified

- `pnpm lint` — clean.
- `pnpm test` — 926 passed, 1 skipped.
- `pnpm translations` — run, diff reviewed.
- E2E suite not run (needs docker); the assertion it depends on is pinned by the unit test above.


elboletaire:
Regarding the "weighted worth a second opinion", a PR with improvements related to this, including re-adding a way to know votings are weighted, is on the way by @ferranrego.

elboletaire:
imatge

PR #1796, opened 2026-09-17
fix(i18n): protect React DOM from browser translation

elboletaire:
claude-gpt-6-astra here, on behalf of elboletaire.

## Findings

- **Verified:** the production renderer served `/pt/admin/processes/create` with `<html lang="en">`, despite a Portuguese title and Portuguese i18next content. The localized SPA catch-all had no `+lang` hook, so vike-react supplied its English default.
- **Verified:** replacing a React-owned text node externally, then inserting a sibling, produces `NotFoundError` with the current React dependency. The route error view renders the exception string without translating it through i18next. Chromium still produces the native exception in English with Portuguese language settings.
- **Inferred:** the reported Portuguese rendering of the native exception strongly supports browser translation as the source of DOM mutation. The exact incident and failing node have not been independently reproduced.
- The production Docker entrypoint serves Vike-generated documents, including the client-rendered admin routes. Changing only the legacy `index.html` template would not protect those documents.

## Changes

- Emit `translate="no"` and `class="notranslate"` on the document through the shared Vike config, before React mounts. This protects SPA and SSR pages, including portal content, from translators that honor the opt-out.
- Set the localized SPA document language from the route and keep it synchronized during native language switches and browser history navigation.
- Add regression tests for document protection, localized route languages, and native language switching/history synchronization.

This intentionally opts the whole app, including public voting pages, out of browser translation. Native i18next translations and the language selector remain available. No DOM-prototype patches, exception suppression, wizard changes, or upstream dependency changes are included.

## Verification

- `pnpm lint` — passed.
- `pnpm test` — 998 passed, 1 existing skipped test.
- `pnpm build` — passed.
- Verified production HTML protection and language metadata for `pt`, `pt-br`, `ca`, and `en` wizard URLs; a public SSR error response also inherits the opt-out.
- Browser smoke checks verified Portuguese/Brazilian Portuguese rendering and preservation of the opt-out after React mounting and theme initialization, without uncaught page errors. Unauthenticated wizard navigation redirected to sign-in as expected.

Actual Google Translate interaction and the authenticated process-creation journey were not replayed. Extensions that ignore translation opt-outs can still mutate the DOM.


elboletaire (commented):
reviewed and checked it works ✅

PR #1799, opened 2026-09-17
fix: resolve the exceptions reported by PostHog error tracking

### LINKED ISSUE #1798: Review and fix posthog notified exceptions.

elboletaire:
This is Opus 5 speaking on behalf of elboletaire.

Reviewed every exception reported by PostHog error tracking over the last 30
days (four distinct issues, all `active`) and fixed each at its root.

## What was reported

| Exception | Occurrences | Where |
|---|---|---|
| `TypeError: 'text/html' is not a valid JavaScript MIME type.` | 12 | public voting domain, root page, Firefox iOS |
| `Error: Script error.` | 15 | same session as the above |
| `UnhandledRejection: Non-Error promise rejection captured with value: Object Not Found Matching Id:1, MethodName:update, ParamCount:4` | 9 | public organization page |
| `TypeError: f.reset is not a function` | 1 | `/es/admin/processes/create` |

## 1. Assets rendered as HTML (`server/`)

The two top issues share one cause. `src/pages/@catchAll/+route.ts` matches
`*`, so an asset the static server could not resolve fell through to the
renderer and was answered with the SPA document. Confirmed against a
production build: a request for a missing chunk returned
`200 Content-Type: text/html`, which is exactly the MIME-type rejection, with
`Script error.` as its opaque `window.onerror` companion.

The same fall-through explains why none of these traces ever symbolicated:
PostHog recorded `Invalid source map: bad json: expected value at line 1
column 1` on every frame, because the `.js.map` fetches were also being
answered with HTML.

Requests whose last path segment carries a known static extension now return
404 with `Cache-Control: no-store`. Matching is by extension rather than an
`/assets/` prefix, so it also covers `public/` files and holds for any `base`.

Vike's `<path>/index.pageContext.json` is rendered on demand rather than read
from disk and is explicitly exempted — 404ing it would silently break
client-side navigation between the SSR public pages. That regression was
caught while verifying and is covered by a test.

## 2. `f.reset is not a function` (`Process/Create`)

react-router types `Blocker.reset` and `Blocker.proceed` as `undefined` in the
`unblocked` and `proceeding` states. `useConfirmOnNavigate` called them
unconditionally. The leave dialog outlives the blocked state — it can still
emit a close after an autosave snoozed the blocker and the effect already
reset it — and `discardAndLeave` turned the throw into a misleading "Error
deleting draft" toast. Since the project builds with `strict: false`, the type
checker never flagged the possibly-undefined call.

Both calls are now optional. The new tests reproduce
`TypeError: blocker.reset is not a function` against the previous code, and
also cover the blocked path so the guard cannot quietly neuter real behaviour.

## 3. Link-scanner noise (`analytics`)

The `Object Not Found Matching Id` rejection is Outlook's Safe Links scanner:
it opens mailed links in a headless Chromium, injects a bridge object and
rejects a promise with a bare string. No stack, no user, and it was a third of
the reported volume. `before_send` now drops it, anchored on the
Id/MethodName/ParamCount triple rather than on "Non-Error promise rejection
captured", which is also how genuine bugs that reject a non-Error arrive — a
test covers that distinction.

## Verification

- `pnpm lint` and `pnpm test` pass (1020 tests, 17 of them new).
- Each behavioural fix was re-run against the unpatched code first to confirm
  the new tests actually fail there.
- The server change was checked against a real production build: missing
  chunks and source maps now 404; existing chunks, `public/` files, SPA
  routes, SSR pages and Vike page-context requests are unchanged. Dev mode was
  smoke-tested too, since the guard sits in a middleware shared by both.

## Deliberately out of scope

- **Chunk-load recovery.** A `vite:preloadError` listener that reloads once
  would recover a user stranded on stale HTML, but that is new resilience
  behaviour with reload-loop risk for everyone, justified by a single session's
  data. Worth a separate change if it recurs.
- **Source-map upload.** Traces will stay minified until maps are built and
  uploaded; the build does not currently emit them, and wiring the upload needs
  a project API key in CI.

Closes #1798


server/static-asset-guard.mjs
```diff
@@ -0,0 +1,98 @@
...
+  // pass `express.static` applies.
+  let decoded
+  try {
+    decoded = decodeURIComponent(pathname)
+  } catch {
+    // Malformed escapes ("%.js") cannot be an app route — their segments are
+    // language codes, hex addresses and base64url ids, none of which carry a
+    // "%". `express.static` does not stop them either: with its default
+    // `fallthrough` it swallows the 400 and calls `next()`, and dev mode never
+    // runs it at all. Fail toward the 404 rather than the renderer.
```
elboletaire: Simplify the comments, three lines max per comment.

src/components/Process/Create/index.test.tsx
```diff
@@ -395,3 +406,302 @@ describe('useFormToVotingProcessRequest', () => {
...
+      const { result } = renderConfirm(blockedBlocker('/admin/processes'))
+
+      expect(result.current.isSamePath).toBe(false)
+    })
+  })
+
+  // The "Save and leave" race: the click's draft write queues behind an
+  // in-flight auto-save, whose completion snoozes the blocker and resets the
+  // pending navigation before `proceed()` runs. `proceed` must then re-issue
+  // the destination itself or the user is silently left on the page.
```
elboletaire: Same, max three lines. This could even be a one-liner, since the test is quite self-descriptive

src/components/Process/Create/index.test.tsx
```diff
@@ -395,3 +406,302 @@ describe('useFormToVotingProcessRequest', () => {
...
+
+      expect(left).toBe(true)
+      expect(paths.at(-1)).toBe('/admin/processes')
+    })
+
+    // `handleSaveAndLeave` awaits its draft write before calling `proceed`, so
+    // the `proceed` it runs was rendered while the blocker was still blocked.
+    // That stale object keeps a `proceed` function, and react-router throws
+    // "Invalid blocker state transition: unblocked -> proceeding" if it is
+    // called after the reset — the fallback has to read the live blocker.
```
elboletaire: simplify

elboletaire (commented):
I've added comments just to some of the comment blocks, review all the new comments added and simplify all of them, making sure none is over three lines.

